PR #13, opened 2015-07-27
keep psf fitter for each obs

beckermr:
This is a simple change to store the fitter for each PSF observation when running the bootstrapper.


beckermr:
We will need this to run the new ngmixer's over the MEDS files.

Issue #5, opened 2014-08-14
Fix the sign error in e1


rmjarvis:
The last run of ngmix still had the sign error in e1.  I believe the only change you need to make is in this line:

https://github.com/esheldon/ngmix/blob/38838c9dc77f9b48ebc6637ea9eba231a0c90209/src/gmix.h#L232

I think this should be (v,u), not (u,v).  i.e. u->col, v->row.  What you have now does u->row, v->col, which means e1 = (Icc-irr)/(Icc+Irr) will be (Ivv-Iuu)/(Ivv+Iuu), which is the source of the sign error.

PR #16, opened 2015-08-05
sign swap for e1

beckermr:
So this looks like the easiest way to swap the sign on e1 addressing issue #5. It def does so, but the value shifts a bit too. I would not accept this change until we have done more testing.

### DIFF
--- a/ngmix/_gmix.h
+++ b/ngmix/_gmix.h
@@ -138,8 +138,8 @@ static double _exp3_lookup[] = {  5.10908903e-12,   1.38879439e-11,   3.77513454
 //#define PYGMIX_MAX_CHI2 1000.0
 
 #define PYGMIX_GAUSS_EVAL_FULL(gauss, rowval, colval) ({            \
-    double _u = (rowval)-(gauss)->row;                         \
-    double _v = (colval)-(gauss)->col;                         \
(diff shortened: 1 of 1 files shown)


esheldon:
I have a branch for this already called fixuv

beckermr:
This PR misses some sign flops in the sums. It looks better to put the changes in _gmix.c like in the fixuv branch. Thus I am closing this.

esheldon:
can you point to an example where yours fails? Fixing the macros does seem the more elegant solution

beckermr:
There are places in the code where you compute direct sums of u and v with the gaussian means. These need to be changed as well.

beckermr:
This line for example 

https://github.com/esheldon/ngmix/blob/master/ngmix/_gmix.c#L1239

esheldon:
ahh, in the EM code.

that is easy to fix.  If we fix that part will the macro take care of the rest?  Looking at my branch I suspect it would

beckermr:
I am not 100% sure. This def explains why the EM PSF fits failed more when I tested my changes.

beckermr:
It is in the round s/n code as well. Not sure where else. Still looking.

beckermr:
It is also here. https://github.com/esheldon/ngmix/blob/master/ngmix/_gmix.c#L2281.

PR #21, opened 2015-09-21
changes to allow for psfex psf normalization

beckermr:
I also fixed bug in passing keywords to the bootstrappers.

### DIFF
--- a/ngmix/bootstrap.py
+++ b/ngmix/bootstrap.py
@@ -550,7 +550,8 @@ class Bootstrapper(object):
                  skip_failed=True,
                  ntry=4,
                  fit_pars=None,
-                 skip_already_done=True):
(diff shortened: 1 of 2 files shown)


esheldon:
These keywords are starting to pile up.  I think we need a more general strategy, and something easier to keep track of, than keywords.  Sounds like a refactor

beckermr:
Maybe. This one is low priority however. 

We could have observations and/or PSFs carry more information about themselves besides images, weights, jacobians and PSFs. Right now I just put that extra info the metadata and tell the function where to get it. 

One proposal would be to add the following attributes for an observation
```python
obs.flux #total flux of an observation
obs.fwhm # FWHM of an observation
```
These could be set to `None` by default.

PR #23, opened 2015-10-09
allow guess widths in max like fits

beckermr:


esheldon:
we need a better way to pass all this around instead of repeating keywords

beckermr:
Agreed. A general cleanup of the code in the bootstrappers might be good with these sorts of configuration parameters in mind.

PR #27, opened 2015-12-09
bug from API change in Shear class

beckermr:


ngmix/guessers.py
```diff
@@ -309,8 +309,7 @@ def get_shape_guess(g1, g2, n, width, max=0.99):
             try:
                 g1_offset = width[0]*srandu()
                 g2_offset = width[1]*srandu()
-                shape_new=shape.copy()
-                shape_new.shear(g1_offset, g2_offset)
+                shape_new = shape.get_sheared(g1_offset, g2_offset)
```
esheldon: let's bike shed a bit here:

do you think this should be make_sheared or get_sheared? I've also added get_sheared to the gmix object, but that is inconsistent with make_image() etc.

beckermr:
Ack. We need ~2 words which say "return a new object w/ this shear applied." "make_sheared" tends to imply that the returned thing is new, but looks funny to me. I think "get_sheared" is a good compromise between descriptive and not funny looking.

beckermr:
Can we merge this or at least decide what to do and fix it?

esheldon:
yeah, let's stick with get_sheared

PR #29, opened 2016-02-05
floating point errors

beckermr:
I have put in some changes to fix floating point errors.

1. In the `C`, for cmodel fits sometimes the `T` normalization can diverge so it checks if `T` is zero anyways and if so, just sets the internal `T` to zero.
2. In the python layer, I made the guessers check before taking logs of negative or zero `T`'s or fluxes. I am clipping the values at `1e-10`. This change may have side effects.


beckermr:
Can we push this into ngmix?

PR #34, opened 2016-03-23
added rendering with fast exp function

beckermr:
This has been fully tested. Here is the code I used to test the functions

```python
#!/usr/bin/env python
import numpy as np
import time
import ngmix

dim=64
jac = ngmix.UnitJacobian(dim/2+1,dim/2+1)
pars = [0.0, 0.0, 0.2, -0.1, 32.0, 100.0]
gmix = ngmix.GMixModel(pars,"exp")
dims=[dim,dim]
ntry = 10

for ns in [1,3,5,10]:
    for npts in [None,5,10]:
        for jc in [jac,None]:
            for FE in [False,True]:

                t0 = time.time()
                for i in xrange(ntry):
                    image = gmix.make_image(dims,jacobian=jc,fast_exp=FE,
                                            nsub=ns,npoints=npts)
                tjac = time.time() - t0

                if FE:
                    print "fast exp:",tjac
                else:
                    print "std exp: ",tjac

                if FE:
                    imfe = image
                else:
                    im = image

            print "max abs diff:",np.max(np.abs(imfe-im))
```

### DIFF
--- a/ngmix/_gmix.c
+++ b/ngmix/_gmix.c
@@ -1016,6 +1016,69 @@ static PyObject * PyGMix_render(PyObject* self, PyObject* args) {
     return Py_None;
 }
 
+/*
+   Render the gmix in the input image, without jacobian
+
+   Error checking should be done in python.
+*/
+static PyObject * PyGMix_render_fast(PyObject* self, PyObject* args) {
+
+    PyObject* gmix_obj=NULL;
+    PyObject* image_obj=NULL;
+    int nsub=0;
+    npy_intp n_gauss=0, n_row=0, n_col=0, row=0, col=0;//, igauss=0;
+    npy_intp rowsub=0, colsub=0;
+
+    struct PyGMix_Gauss2D *gmix=NULL;
+    double *ptr=NULL, stepsize=0, offset=0, areafac=0, tval=0, trow=0, tcol=0;
+
+    if (!PyArg_ParseTuple(args, (char*)"OOi", &gmix_obj, &image_obj, &nsub)) {
+        return NULL;
+    }
+
+    gmix=(struct PyGMix_Gauss2D* ) PyArray_DATA(gmix_obj);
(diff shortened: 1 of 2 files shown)


esheldon:
Wow, that's a mess of logic!

beckermr:
No better way to do it...

beckermr:
You get max abs differences around 1e-6 which I think is about right for the apprx exp function.

esheldon:
There is a huge amount of code duplication here.  We should spend a few minutes thinking if we can re-use code and retain the speedup.

esheldon:
max abs difference in a pixel?  The fast exp should be good to a part in 10^15

esheldon:
oh, except if the arg exceeds 25, in which case it is not evaluated....

esheldon:
your code generation idea might be the best way to simplify this.

esheldon:
also, the nsub way of integrating over pixels is obsolete, we can do away with it

beckermr:
these are all good suggestions but I don't have time to do this

the easy way to do the code generation would be to define the ngmix functions in an include in C without include guards. then you can define a few macros before the include and build both sets.

beckermr:
It would be great we could accept this PR now and make an issue to fix this later.

esheldon:
I've seen that done, but I don't actually like it, because it means the same code means different things in different contexts

beckermr:
you can redefine the function names too using string concatenation with macros

esheldon:
I'm willing to accept it now, since time is short for you and we need to get a larger run of teh code done on real data.

esheldon:
argh, this is against master so the merge with the branch metacalwcs will be a pain

beckermr:
frack

beckermr:
revert if you need to

beckermr:
I can push to another branch, but merge will be a pain no matter what

PR #42, opened 2018-09-14
BUG change the init for fracdev to let the sims converge

beckermr:
See my branch on my repo.

@NiallMac for viz

### DIFF
--- a/ngmix/bootstrap.py
+++ b/ngmix/bootstrap.py
@@ -57,7 +57,7 @@ class Bootstrapper(object):
         obs: observation(s)
             Either an Observation, ObsList, or MultiBandObsList The
             Observations must have a psf set.
-            
+
             If the psf observations already have gmix objects set, there is no
             need to run fit_psfs()
         """
@@ -279,7 +279,7 @@ class Bootstrapper(object):
 
             psf_T=psf_T_sum/wsum
 
-            # and finally, do the fit 
+            # and finally, do the fit
             try:
                 round_fitter=self._fit_sim_round(fitter,
                                                  pars_round,
@@ -290,7 +290,7 @@ class Bootstrapper(object):
                 res=round_fitter.get_result()
                 if res['flags'] != 0:
                     print("        round fit fail")
-                    flags |= BOOT_TS2N_ROUND_FAIL 
+                    flags |= BOOT_TS2N_ROUND_FAIL
                 else:
                     import covmatrix
                     try:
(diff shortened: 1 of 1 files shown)


esheldon:
How generic do you think this is?  Should we also adopt this for real data runs with deeper data for example? Is there something particular about the sims that is causing the problem?

beckermr:
I don't know. Basically the code was starting at 1.5 which is the max for the old guessing array. The whole range of allowable fracdev values is tough. Allowing them outside of [0, 1] generally gives better fits IIRC but is unphysical at some level. 

I think we could try a small test patch in the data and see what happens.

beckermr:
@esheldon I vote to merge this and deal with the consequences. We will never be able to properly test it all cases. :doitlive: :shipit:

esheldon:
can we make it configurable, a keyword or part of a fracdev guesser or some such?

esheldon:
also might be better to start from master since this is going to conflict

ngmix/bootstrap.py
```diff
@@ -1840,7 +1840,8 @@ def __init__(self, obs,
                                         fracdev_grid['max'],
```
beckermr: @esheldon It may already be configurable. IDK where `fracdev_grid` comes from.

beckermr: If we can just change the config, then we should close this PR and move on.

beckermr:
The merge conflicts are NBD.

ngmix/bootstrap.py
```diff
@@ -1840,7 +1840,8 @@ def __init__(self, obs,
...
                                         fracdev_grid['num'])
         else:
-            self.fracdev_tests=linspace(-1.0,1.5,26)
+            # self.fracdev_tests=linspace(-1.0,1.5,26)
+            self.fracdev_tests=linspace(0.0,1.0,26)
```
beckermr: @esheldon here!

esheldon: ok I understand now

esheldon: you can input the fracdev grid parameters, but this is the default.  So better to add something to the config file and then make ngmixer send it on to this bootstrapper

esheldon: agreed we should update the default too though

esheldon:
do you mind if I start a new PR with just this change to avoid the clutter from the whitespace changes?

ngmix/bootstrap.py
beckermr: ngmixer just sends everything straight through. It might already work. Let me check.

beckermr: yeah ngmixer just sends it all straight in.

EG, https://github.com/esheldon/ngmixer/blob/master/ngmixer/bootfit.py#L786

beckermr:
Shall we close this? I think the old code is old and we should move on.

beckermr:
Closing. Feel free to reopen.

PR #44, opened 2018-11-03
ENH add meds readers for ngmix obs

beckermr:
This PR has MEDS readers subclassed to output ngmix observations. I think this is the right spot for them as opposed to the MEDS repo (i.e., if you need the ngmix stuff, then you have ngmix installed).

### DIFF
--- a/ngmix/medsreaders.py
+++ b/ngmix/medsreaders.py
@@ -1,5 +1,5 @@
 import numpy as np
-import meds
+from meds import MEDS as _MEDS
 
 from .observation import MultiBandObsList, Observation, ObsList
 from .jacobian import Jacobian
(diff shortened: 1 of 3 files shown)


ngmix/medsreaders.py
```diff
@@ -101,7 +101,7 @@ def get_mbobs(self, iobj, weight_type='weight'):
         return mbobs
 
 
-class MEDS(meds.MEDS):
+class MEDS(_MEDS):
```
esheldon: Shall we rename this class to something distinct from the original class name?  e.g. `NGmixMeds`?

beckermr: sure!

beckermr:
@esheldon renamed some stuff. Ready for a second look!

PR #50, opened 2018-12-11
ENH enable user to get only the requested types for metacal

beckermr:
This is a small change to allow ppl to get only one of the metacal types if they wish. It preserves the old behavior.

### DIFF
--- a/ngmix/metacal.py
+++ b/ngmix/metacal.py
@@ -298,6 +298,9 @@ class Metacal(object):
             The shear step value to use for metacal
         types: list
             Types to get.  Default is given in METACAL_TYPES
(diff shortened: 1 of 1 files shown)


beckermr:
@esheldon Lets not merge this quite yet. I am still testing locally.

esheldon:
probably should rewrite this to do it properly:  just return exactly the ones that are requested, and only if none are sent should the default set be used

beckermr:
I didn't want to break existing code, which is why the PR looks the way it does. I am happy to make the change to the proper way, but I am a bit scared...

esheldon:
I'm going to do a slightly different PR to address this

PR #52, opened 2019-01-08
better handling of types= in get_all for metacal

esheldon:
If no types are sent, default to METACAL_TYPES

no longer enforce required types

Also make step a keyword and default to 0.01

Added a unit test for get_all_metacal to make sure
things are working


esheldon:
@beckermr

beckermr:
Are we good to merge this? I am going to be making a PR against it.

beckermr:
Huzzah!

Issue #56, opened 2019-02-08
"TypeError" when doing metacal+shape measurement with bootstrap


aguinot:
Hello,

I tried running the metacal example (using the bootstrap solution) from the Wiki page and I got the following error on a lot of images :
```python
TypeError: __init__() missing 1 required positional argument: 'value'
```
This error seems to come from `scipy.optimize.leastsq` function..

Here are the version of the different libraries used :
python = 3.6.8
scipy = 1.2.0
ngmix = v1.2 (clone from the master branch)
galsim = 2.1.4

esheldon:
This is due to a bug in the latest release of numba.

You can temporarily fix this by downgrading the numba version using `conda install numba=0.40`

aguinot:
Thank you for the quick answer, this fix work for me.
I guess you can cancelled my pull request.

esheldon:
I will leave it open for now, so others can see the temporary fix. I'll close it when the next version of numba is released (which should have a fix)

PR #57, opened 2019-02-08
Fix to TypeError

aguinot:
Possible fix concerning the issue #56


esheldon:
I think most people have probably moved off that old version of numba. I'm going to close this.

PR #66, opened 2019-07-09
use psf_cutout_{row,col} for psf cen in psf obs

esheldon:


esheldon:
I ran some tests and the center of the fit ends up quite close to the psf cutout row/col

PR #76, opened 2019-08-29
TST added tests for observations

beckermr:
This PR adds tests for `Observation` and the various lists.


beckermr:
There are no changes to the main source code here. Merging.

PR #78, opened 2019-08-30
Psfflux offset

esheldon:
fixing psf offset in galsim psf flux fitter


esheldon:
I bumped the version here so my codes see a different version string, but no release is planned yet

PR #82, opened 2019-08-30
forgot to add the actual test

esheldon:


esheldon:
yes, but I don't have time right now and need to make a release

PR #94, opened 2019-09-17
Update pixels zero

esheldon:
added constructor option to Observation about ignoring the zero weight pixels

default is still to ignore, but on can now specify `ignore_zero_weight=False`.

Added a test.

Also threw in returning the correct sky in em.prep_sky

### DIFF
--- a/ngmix/observation.py
+++ b/ngmix/observation.py
@@ -7,7 +7,8 @@ import copy
 
 from .pixels import make_pixels
 
-DEFAULT_XINTERP='lanczos15'
+DEFAULT_XINTERP = 'lanczos15'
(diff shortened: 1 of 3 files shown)


ngmix/observation.py
```diff
@@ -45,7 +45,10 @@ def __init__(self,
                  jacobian=None,
                  gmix=None,
                  psf=None,
-                 meta=None):
+                 meta=None,
+                 ignore_zero_weight=True):
+
+        self._ignore_zero_weight = ignore_zero_weight
```
beckermr: You need a doc string entry for this option.

PR #97, opened 2019-10-03
Introduce new approximate exp functions

esheldon:
I introduce a 4th and 5th order exp approximate function, as well as the old 3rd order

```
exp3
exp4
exp5
```

We are using exp5 but I leave the others here for now, we can remove later.

I also altered the tests to more lenient. Basically as long as we are doing fits to a part in 10 to 5th we are certainly happy and the tests all pass with that tolerance

The new exp functions are limited to the range over which we actually use them.  We always use a max chi squared of 25, so the range of the lookup table need only be from about -15 to zero

This also means our max chi squared is hard coded, so I took out the max_chi2 keywords, which were always set to 25 anyway


esheldon:
my commit comments are a little misleading

by "target absolute errors" I just mean don't compare to errors since there will definitely be bias more than errors in many cases.  We just want it to be pretty small, and 1.0e-5 seems to be good enough

esheldon:
again sorry, I mean "dont' compare to uncertainties"

ngmix/em_nb.py
```diff
@@ -113,8 +113,8 @@ def do_scratch_sums(pixel, gmix, sums):
         chi2 = \
             gauss['dcc']*v2 + gauss['drr']*u2 - 2.0*gauss['drc']*uv
 
-        if chi2 < 25.0 and chi2 >= 0.0:
-            tsums['gi'] = gauss['pnorm']*expd( -0.5*chi2 )
+        if chi2 < FASTEX_MAX_CHI2 and chi2 >= 0.0:
+            tsums['gi'] = gauss['pnorm']*exp5(-0.5*chi2)
```
beckermr: Should we hide this def behind a simple global name? 

```
from .fastexp import fexp, FASTEXP_MAX_CHI2
...
```

where `fexp` is

```
if True:
    fexp = exp5
elif False:
    fexp = exp3
```

etc.

ngmix/em_nb.py
```diff
@@ -12,7 +12,7 @@
     gauss2d_set_norm,
     gmix_get_e1e2T,
 )
-from .fastexp import expd
+from .fastexp import exp5, FASTEX_MAX_CHI2
```
beckermr: Did we want to make this `FASTEXP_MAX_CHI2` (notice the `P`)?

ngmix/tests/test_fastexp.py
```diff
@@ -2,19 +2,23 @@
 import numpy as np
 import pytest
 
-from ngmix.fastexp import expd
+from ngmix.fastexp import exp5
 
+# test values between -15 and 0
+vals = [-7.8864744,  -4.2333561, -11.02660361,  -9.07802778,
+        -12.01531878,  -8.4256256,  -8.70588303]
```
beckermr: Something near zero might be good too?

ngmix/em_nb.py
esheldon: I actually like seeing the exp5 there for now, but I agree we should maybe rename later

ngmix/em_nb.py
esheldon: good catch

beckermr:
I fixed flake8.

PR #98, opened 2019-10-07
PSF fitting features

esheldon:
I accidentally added multiple different things into one commit.

In addition to higher ngauss coellip models, also added sending a guess to the psf runners

### DIFF
--- a/ngmix/bootstrap.py
+++ b/ngmix/bootstrap.py
@@ -2011,22 +2011,31 @@ class PSFRunner(object):
         self.lm_pars=lm_pars
         self.set_guess0(Tguess)
 
-    def go(self, ntry=1):
+    def go(self, ntry=1, guess=None):
+
         from .fitting import LMSimple
 
         for i in xrange(ntry):
 
-            guess=self.get_guess()
+            if i == 0 and guess is not None:
+                this_guess = guess.copy()
+            else:
+                this_guess = self.get_guess()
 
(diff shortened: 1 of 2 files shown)


ngmix/bootstrap.py
```diff
@@ -2430,6 +2461,13 @@ def set_rng(self, rng):
 
 _moffat4_pguess=array([0.44534,  0.366951,  0.10506,  0.0826497])
 _moffat4_fguess=array([0.541019,  1.19701,  0.282176,  3.51086])
+
+_moffat5_pguess=array([0.45,  0.25,  0.15,  0.1, 0.05])
+_moffat5_fguess=array([0.541019,  1.19701,  0.282176,  3.51086, ])
+
+_moffat5_pguess = array([0.57874897, 0.32273483, 0.03327272, 0.0341253 , 0.03111819])
+_moffat5_fguess = array([0.27831284, 0.9959897 , 5.86989779, 5.63590429, 4.17285878])
+
```
beckermr: We should add some tests that these numbers all sum properly to 1 or whatever that condition is.

esheldon: we don't have those for the other ones, we'd have to do them all

I suggest we do that in a separate PR

esheldon: also note that if they are not properly normalized, it won't actually matter

esheldon: (these are just for guesses in the fitters)

beckermr: sounds good!

PR #102, opened 2020-03-16
bug fix not updating obs pixel array in metacal when adding noise

esheldon:


beckermr:
Wait, does this mean that the image pixels were not being updated as well?

beckermr:
Ohhhh. It's only indexing. Yikes. We should change the API to not allow this or fix this. That is a tough corner case to remember.

beckermr:
LGTM!

esheldon:
No, it isn't just indexing.

I did the image correctly, or none of the sims would have worked
```
obs.image = im + nim
```
but for the weight I did it in place
```
obs.weight[wpos] = 1.0/tvar[wpos]
```

beckermr:
That is what I am saying. When you modify the image by assignment it is fine and the bug only happens for indexing operations.

esheldon:
Yes.

We do want the user to be able to get a reference to the image though.  I'm not sure how to prevent it being modified

beckermr:
Yeah me neither, but this bug is scary. 

Maybe we introduce a context manager on the pixels data structure that forces an update when entering. This update should be fast. Then we can use the context manager whenever we access the pixels.

esheldon:
for the regular image property we should return a view with the read only flag set.

The context sounds like a nice way to implement doing modifications.  e.g.

```python
im = obs.image
im[w] = 35  # raises error due to read only flag

with obs.modifying() as modobs:
    im = modobs.image  # no read only flag set
    wt  modobs.weight  # no read only flag set
    # do some modifyin

# on exit of the context, update_pixels() is run
```

esheldon:
(not sure how to actually do the above correctly)

beckermr:
I can make a PR but I don't have the time.

beckermr:
For now, we could break the API and set a ready only flag on all attributes via numpy. Then we can add back the context manager in the future. Note that this may break a lof of our code, but would make any potential bugs obvious (raising a `RuntimeError`).

One detail here. We should make a copy of the inputs before doing this, since any existing references to the object or a slice of it may not inherit this change and thus may not get the flag. 
See the doc string comments and text around WRITEABLE here: https://docs.scipy.org/doc/numpy/reference/generated/numpy.ndarray.flags.html

esheldon:
```python
import numpy as np


class C(object):
    def __init__(self):
        self._data = np.zeros(5)
        self._writeable = False

    @property
    def data(self):
        dview = self._data.view()
        dview.flags['WRITEABLE'] = self._writeable
        return dview

    def writeable(self):
        self._writeable = True
        return self

    def __enter__(self):
        return self

    def __exit__(self, exception_type, exception_value, traceback):
        self._writeable = False


c = C()
print(c.data)

try:
    c.data[1] = 5
except ValueError as err:
    print("as expected raised error '%s'" % err)


with c.writeable() as cmod:
    cmod.data[1] = 5

print(c.data)

try:
    c.data[1] = 5
except ValueError as err:
    print("as expected raised error '%s'" % err)

```

esheldon:
that's not quite right because it can be used outside a context

beckermr:
The docs say a RuntimeError. Weird.

People could still muck with `_image` here, but I guess that is on them.

beckermr:
> that's not quite right because it can be used outside a context

example?

esheldon:
With the above, one could just do this at any time, outside a context
```
c.writeable()
```

esheldon:
but as you say, if someone did that it would be on them.

beckermr:
Ahh hmm. Idk how to solve that. We should make writable a ready only property in any case.

PR #104, opened 2020-03-16
Implement writeable context for observations

### LINKED ISSUE #103: make image, weight etc. refs of observations read only by default
add a context manger that switches on writeable refs and runs update_pixels on exit

this brings up again the issue of jacobians, which are always copies in the current code

esheldon:
closes #103 

Update the metacal addnoise code to use the context


esheldon:
I don't think we require it for the other images, they are not stored in the pixels array

beckermr:
Sure but the api is pretty fractured then.

esheldon:
after more thought I agree.

But while we are at it we should fix the inconsistency of the jacobian, which returns a copy.  This will definitely break some code

esheldon:
ok to merge?

ngmix/bootstrap.py
```diff
@@ -279,93 +274,6 @@ def _get_round_pars(self, pars_in):
 
         return pars, pars_lin
 
-
-    def _find_cen(self, ntry=10, Tguess=4.0):
```
beckermr: What happened to this method? Do we not need it anymore?

esheldon: no, it was never used in production

PR #117, opened 2020-09-21
TST added tests for observations

beckermr:
This PR catches some edge cases missed in the current tests.

Many of the tests here are simple sanity checks. We have or will have functional tests in the fitting routines which depend on much of the functionality in the classes here.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,5 @@
+[run]
+omit =
+    # we omit numba code since we cannot measure it
+    ngmix/*_nb.py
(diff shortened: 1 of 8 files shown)


esheldon (approved):
Thanks Matt.  I think there are more tests we could add, but these look good.

beckermr:
> Thanks Matt. I think there are more tests we could add, but these look good.

The coverage is at nearly 100% and I think I captured all of the major exceptions and branches. What else were you thinking?

PR #122, opened 2020-10-19
ENH add tests and docs for priors

beckermr:
This PR has tests and docs for the `priors` module.


ngmix/priors.py
```diff
@@ -445,16 +625,12 @@ def get_fdiff_scalar(self, g1, g2):
...
-        is on |g|, so the sign doesn't matter
-        """
-
+    def _get_fdiff_array(self, g1, g2):
+        # For use with LM fitter, which requires
+        #     (model-data)/width
+        # so we need to fake it
+        # In this case the fake fdiff works OK because the prior
+        # is on |g|, so the sign doesn't matter
         lnp = self.get_lnprob_scalar2d(g1, g2)
```
beckermr, line 634: I think I'm tired but I suspect this code never worked with arrays.

esheldon: I think you're right

esheldon:
I think we can remove these

```python
GPriorGreat3Exp
GPriorGreatDES
GPriorGreatDES2
GPriorGreatDESNoExp
make_gprior_great3_exp
make_gprior_great3_bdf
make_gprior_cosmos_exp
make_gprior_cosmos_dev
make_gprior_cosmos_sersic
GPriorM
_gprior2d_exp_scalar
_gprior2d_exp_array
_gprior1d_exp_scalar
GPriorMErf
GPriorMErf2
FlatEtaPrior
lognorm_convert_old
BFracBase
BFrac
JOINT_N_ITER
JOINT_MIN_COVAR
JOINT_COVARIANCE_TYPE
TFluxPriorCosmosBase
TFluxPriorCosmosExp
TFluxPriorCosmosDev
TFluxPriorCosmosBase
GPriorCosmosSersicSpline
_g_cosmos_t
_g_cosmos_c
_g_cosmos_k
```

esheldon:
Feeling lean :)

What does the DEP mean?

beckermr:
I used it to mean deprecate, but actually, this is removal.

beckermr:
https://numpy.org/doc/1.16/dev/gitwash/development_workflow.html#writing-the-commit-message

beckermr:
OK @esheldon this PR fixes two bugs and brings test coverage to ~55% from ~49%. I will be working on the rest of the tests for the rest of the code here is another PR. Ready for review and merge!

ngmix/priors/__init__.py
```diff
@@ -0,0 +1,3 @@
+# flake8: noqa
+from .priors import *
+from .random import make_rng, srandu
```
esheldon, line 3: It would be better to remove srandu entirely and instead make the functions that call it *require* having an rng sent.  This stems from the times when I used the default rng from numpy

ngmix/priors/random.py
```diff
@@ -0,0 +1,27 @@
...
+    but with a seed generated from current numpy state
+    """
+    if rng is None:
+        seed = numpy.random.randint(0, 2 ** 30)
+        rng = numpy.random.RandomState(seed)
+
+    return rng
+
+
+def srandu(num=None, rng=None):
```
esheldon, line 16: should remove and require the functions that use it to use the sent rng, and require that rng is not None

esheldon (commented):
I think the most important test for these and upcoming PRs is to actually run the lsst sims with metadetect to see if it gives the same answers as before.

we should dump srandu, either in this PR or in a followup

beckermr:
> I think the most important test for these and upcoming PRs is to actually run the lsst sims with metadetect to see if it gives the same answers as before.

Agreed but I had hope that the small changes and bug fixes in these low-level routines would be ok. How often do you want to do this check?

esheldon:
I agree should be OK, just making a general statement.  We should test soon.

beckermr:
OK. I made rng keyword required for `srandu`. Note that I shimmed back in the global numpy RNG in places to keep the code behavior as close as possible to the same as it is now.

esheldon:
That stuff with the guessers is exactly what I was thinking we should dump, should never use the default rng.  but agree let's do it in a different PR

esheldon (approved):

PR #127, opened 2021-01-18
proposal for fitting api

esheldon:
this is a proposal for a new fitting/runner/bootstrap api

a key aspect is that most of the code is basically "pure" in the sense that the internal state of the input data (observations) are not changed.

The bootstrapping (currently no metacal) is just a function that runs runners.  It is there were some mutation occcurs, setting psf mixtures that are needed by the galaxy fitters.

The classes for fitters and runners only have one required method by the api "go(...)" but of course others may be needed such as `get_gmix` for the psf fitting

### DIFF
--- a/proposals/fitapy_proposal.py
+++ b/proposals/fitapy_proposal.py
@@ -1,0 +1,462 @@
+"""
+schematic proposal for a new api for fitters, runners, and bootstrapping
+
+bootstrapping is now just a function that runs the runners that are
+input
+"""
+from .observation import (
+    Observation,
+    ObsList,
+    MultiBandObsList,
+)
+from .gexceptions import BootPSFFailure
+
+
+def bootstrap(
+    *,
+    obs,
+    runner,
+    psf_runner,
+    remove_failed_psf=True,
+):
+    """
+    Run a fitter on the input observations, possibly bootstrapping the fit
+    based on information inferred from the data or the psf model
+
+    Parameters
+    ----------
+    obs: ngmix Observation(s)
+        Observation, ObsList, or MultiBandObsList
+    runner: ngmix Runner
+        A runner or measurement to run the fit.
+    psf_runner: ngmix PSFRunner
+        A runner to run the psf fits
+    remove_failed_psf: bool, optional
+        If set to True, remove observations where the psf fit fails, and
+        only fit the remaining.  Default True.
+
+    Returns
+    -------
+    result dict
+
+    Side effects
+    ------------
+    the psf.meta['result'] and the .psf.gmix may be set if a psf runner is sent
+    and the internal fitter has a get_gmix method.  gmix are only set for
+    successful fits
+
+    Note the only place where side effects should be occurring is
+    in the call to store_psf_results.
+    """
+
+    if psf_runner is not None:
+        psf_result = psf_runner.go(obs=obs)
+
+        if remove_failed_psf:
+            obs, psf_result = remove_failed_psf_obs(
+                obs=obs, result=psf_result,
+            )
+
+        if result_has_gmix(psf_result):
+            set_gmix = True
+        else:
+            set_gmix = False
+
+        store_psf_results(
+            obs=obs, result=psf_result, set_gmix=set_gmix,
+        )
+
+    return runner.go(obs=obs)
+
+
+class MeasurementBase(object):
+    """
+    Perform measurements or fits on observations.  The only required
+    method is go()
+    """
+    def go(self, *, obs):
+        """
+        Run a measurement or fit on the input observation(s)
+
+        Parameters
+        ----------
+        obs: ngmix Observation(s)
+            Observation, ObsList, or MultiBandObsList
+
+        Returns
+        -------
+        None
+
+        Side Effects
+        -------------
+        the result attribute is set
+        """
+        raise NotImplementedError('implement go method for fitter')
+
+
+class Runner(MeasurementBase):
+    """
+    Run a fitter and guesser on observations
+
+    Parameters
+    ----------
+    fitter: ngmix fitter or measurer
+        An object to perform measurements, must have a go(obs=obs, ntry=ntry)
+        method.
+    guesser: ngmix guesser object
+        Must be a callable returning an array of parameters
+    ntry: int, optional
+        Number of times to try if there is failure
+
+    Side effects
+    ------------
+    None
+    """
+    def __init__(self, *, fitter, guesser, ntry=1):
+        self.fitter = fitter
+        self.guesser = guesser
+        self.ntry = ntry
+
+    def go(self, *, obs):
+        """
+        Run the fitter on the input observation(s), possibly multiple times
+        using guesses generated from the guesser
+
+        Parameters
+        ----------
+        obs: ngmix Observation(s)
+            Observation, ObsList, or MultiBandObsList
+
+        Returns
+        -------
+        result dictionary
+        """
+
+        return run_fitter(
+            obs=obs, fitter=self.fitter, guesser=self.guesser, ntry=self.ntry,
+        )
+
+
+class PSFRunner(Runner):
+    """
+    Run a fitter on each psf observations.
+
+    Parameters
+    ----------
+    fitter: ngmix fitter or measurer
+        An object to perform measurements, must have a go(obs=obs, ntry=ntry)
+        method.
+    guesser: ngmix guesser object
+        Must be a callable returning an array of parameters
+    ntry: int, optional
+        Number of times to try if there is failure
+    """
+
+    def go(self, *, obs):
+        """
+        Run the fitter on the psf observations associated with the input
+        observation(s), possibly multiple times using guesses generated from
+        the guesser.
+
+        Parameters
+        ----------
+        obs: ngmix Observation(s)
+            Observation, ObsList, or MultiBandObsList
+        ntry: int, optional
+            Number of times to try if there is failure
+
+        Returns
+        -------
+        result dict, or list of, or list of list of, depending on the
+        input type
+
+        Side effects
+        ------------
+        None
+        """
+
+        return run_psf_fitter(
+            obs=obs, fitter=self.fitter, guesser=self.guesser, ntry=self.ntry,
+        )
+
+
+class MaxlikeFitter(MeasurementBase):
+    """
+    Perform maximum likelihood fits on observations
+
+    Parameters
+    ----------
+    prior: ngmix prior, optional
+        A prior to use for the fit
+    """
+
+    def __init__(self, *, prior=None):
+        self.prior = prior
+
+    def go(self, *, obs, guess):
+        """
+        Run a measurement or fit on the input observation(s)
+
+        Parameters
+        ----------
+        obs: ngmix Observation(s)
+            Observation, ObsList, or MultiBandObsList
+        guess: array-like
+            Initial starting point for the fit
+
+        Returns
+        -------
+        result dict
+
+        Side effects
+        ------------
+        None
+        """
+        raise NotImplementedError('implement go method for fitter')
+
+    def get_gmix(self, *, pars):
+        raise NotImplementedError('implement get_gmix method for fitter')
+
+
+class Moments(MeasurementBase):
+    """
+    Measure moments from observations
+
+    Parameters
+    ----------
+    weight: gaussian mixture object
+        The weight to use for the measurements
+    """
+
+    def __init__(self, *, weight):
+        self.weight = weight
+
+    def go(self, *, obs):
+        """
+        Measure weighted moments on the observation(s)
+
+        Parameters
+        ----------
+        obs: ngmix Observation(s)
+            Observation, ObsList, or MultiBandObsList
+
+        Returns
+        -------
+        result dict
+
+        Side effects
+        ------------
+        None
+        """
+        raise NotImplementedError('implement go method for fitter')
+
+
+def remove_failed_psf_obs(*, obs, result):
+    """
+    remove observations from the input that failed
+
+    Parameters
+    ----------
+    obs: observation(s)
+        Observation, ObsList, or MultiBandObsList
+    result: result dict, or list or list of lists
+        The results for the psf fits
+
+    Returns
+    --------
+    obs, result:
+        new observations, same type as input, as well
+        as new results
+    """
+    if isinstance(obs, MultiBandObsList):
+        new_mbobs = MultiBandObsList(meta=obs.meta)
+        new_mbobs_result = []
+        for tobslist, treslist in zip(obs, result):
+
+            new_obslist = ObsList(meta=tobslist.meta)
+            new_obslist_result = []
+            for tobs, tres in zip(tobslist, treslist):
+                if tres['flags'] == 0:
+                    new_obslist.append(tobs)
+                    new_obslist_result.append(tres)
+
+            if len(new_obslist) == 0:
+                raise BootPSFFailure('no good psf fits')
+
+            new_mbobs.append(new_obslist)
+            new_mbobs_result.append(new_obslist_result)
+
+        return new_mbobs, new_mbobs_result
+
+    elif isinstance(obs, ObsList):
+        new_obslist = ObsList(meta=tobslist.meta)
+        new_obslist_result = []
+        for tobs, tres in zip(obs, result):
+            if tres['flags'] == 0:
+                new_obslist.append(tobs)
+                new_obslist_result.append(tres)
+
+        if len(tobslist) == 0:
+            raise BootPSFFailure('no good psf fits')
+
+        return new_obslist, new_obslist_result
+    else:
+        if result['flags'] != 0:
+            raise BootPSFFailure('no good psf fits')
+
+        return obs, result
+
+
+def run_fitter(*, obs, fitter, guesser, ntry=1):
+    """
+    run a fitter multiple times if needed, with guesses generated from the
+    input guesser
+
+    Parameters
+    ----------
+    obs: ngmix Observation(s)
+        Observation, ObsList, or MultiBandObsList
+    fitter: ngmix fitter or measurer
+        An object to perform measurements, must have a go(obs=obs, ntry=ntry)
+        method.
+    guesser: ngmix guesser object
+        Must be a callable returning an array of parameters
+    ntry: int, optional
+        Number of times to try if there is failure
+
+    Returns
+    -------
+    result dictionary
+    """
+
+    for i in range(ntry):
+
+        guess = guesser()
+        res = fitter.go(obs=obs, guess=guess)
+
+        if res['flags'] == 0:
+            break
+
+    return res
+
+
+def run_psf_fitter(*, obs, fitter, guesser, ntry=1):
+    """
+    run a fitter on each psf observation in the input observation(s).  The
+    fitter will be run multiple times if needed, with guesses generated from
+    the input guesser
+
+    Parameters
+    ----------
+    obs: ngmix Observation(s)
+        Observation, ObsList, or MultiBandObsList
+    fitter: ngmix fitter or measurer
+        An object to perform measurements, must have a go(obs=obs, ntry=ntry)
+        method.
+    guesser: ngmix guesser object
+        Must be a callable returning an array of parameters
+    ntry: int, optional
+        Number of times to try if there is failure
+
+    Returns
+    -------
+    result dict, or list of, or list of list of, depending on the
+    input type
+
+    Side Effects
+    ------------
+    .meta['result'] is set to the fit result and the .gmix attribuite is set
+    for each successful fit
+    """
+
+    if isinstance(obs, MultiBandObsList):
+        result = []
+        for tobslist in obs:
+            tres = run_psf_fitter(
+                obs=tobslist, fitter=fitter, guesser=guesser, ntry=ntry,
+            )
+            result.append(tres)
+
+    elif isinstance(obs, ObsList):
+        result = []
+        for tobs in obs:
+            tres = run_psf_fitter(
+                obs=tobs, fitter=fitter, guesser=guesser, ntry=ntry,
+            )
+            result.append(tres)
+
+    elif isinstance(obs, Observation):
+        result = run_fitter(
+            obs=obs.psf, fitter=fitter, guesser=guesser, ntry=ntry,
+        )
+        if result['flags'] == 0 and hasattr(fitter, 'get_gmix'):
+            result['gmix'] = fitter.get_gmix(result['pars'])
+    else:
+        raise ValueError(
+            'obs must be an Observation, ObsList, or MultiBandObsList'
+        )
+
+    return result
+
+
+def result_has_gmix(result):
+    """
+    check if a gmix is present.  Not all are checked
+    """
+    try:
+        if 'gmix' in result:
+            return True
+        else:
+            return False
+    except TypeError:
+        return result_has_gmix(result[0])
+
+
+def store_psf_results(*, obs, result, set_gmix=True):
+    """
+    store the psf results in the obs meta and possibly set the gmix
+    for the psf
+
+    Parameters
+    ----------
+    obs: ngmix Observation(s)
+        Observation, ObsList, or MultiBandObsList
+    result: result dict, list or list of lists
+        The results to store
+    set_gmix: bool, optional
+        If set to True, the gmix is constructed for the psf and
+        set
+
+    Returns
+    -------
+    None, the result will be stored in the meta for each observation
+
+    Side Effects
+    ------------
+    .meta['result'] is set to the fit result and the psf.gmix attribuite is set
+    for each successful psf fit
+    """
+
+    if isinstance(obs, MultiBandObsList):
+        for tobslist, tresult in zip(obs, result):
+            store_psf_results(obs=tobslist, result=tresult, set_gmix=set_gmix)
+
+    elif isinstance(obs, ObsList):
+        for tobs, tresult in zip(obs, result):
+            store_psf_results(obs=tobs, result=tresult, set_gmix=set_gmix)
+
+    elif isinstance(obs, Observation):
+        if result['flags'] != 0:
+            raise ValueError('cannot set gmix for failed fit')
+
+        obs.psf.meta['result'] = result
+        if set_gmix:
+            if 'gmix' not in result:
+                raise ValueError('result has no gmix item')
+
+            obs.psf.gmix = result['gmix']
+
+    else:
+        raise ValueError(
+            'obs must be an Observation, ObsList, or MultiBandObsList'
+        )


esheldon:
This requires a change to the fitters to return a result dict.

Note that the constructor of the fitters does not take the observation any more, rather they can operate on observations with their go methods.  This is a big difference

ngmix/gaussmom.py
```diff
@@ -48,56 +43,7 @@ def get_result(self):
 
         return self.result
 
-    def _do_coadd_maybe(self, mbobs):
```
beckermr, line 51: This is a big API change. Is there a reason we don't want to allow this?

esheldon: I never used this code gaussmom code anywhere, it was just a sort of copy of the moments stuff from metadetect

esheldon: I think there is no purpose to actually doing this coadding here anyway

esheldon:
I guess this means the tests from master are run rather than the tests from this branch

esheldon:
Questions about priors:

- Do we want to make rng a required parameter?  I can't think of an actual use case where I would not want control over that.  Its buggy not making it required:  I noticed that we hadn't sent the rng to the priors in some of the prior tests, which made the tests not repeatable.  Its easy to forget
- Do we need `MultivariateLogNormal` and `MVNMom` or `TruncatedSimpleGauss2D`?

esheldon:
to be clear:  the rng is used for sampling.  the sampling is used most often in guessers: either a sample is used for part of the guess or samples are used to "fix" guesses that go out of bounds of the prior

beckermr:
Yes. I think we should make the RNG required. 

I don't know of any uses of the multivariate distributions. If you think they are extraneous, then we should get rid of them.

esheldon:
OK.   We can always add them back in if it turns out they were used somewhere

CHANGES.md
```diff
@@ -1,5 +1,16 @@
 ## unreleased (v2.0.0)
 
+### breaking API changes
+
+- The fitters no longer take the observation in the constructor.
+  The `go` method now takes the observation and possibly a guess.
```
beckermr, line 6: Did we want to enable the fitters to take a list of mbobs?

this is what is in the mdet repo: https://github.com/esheldon/metadetect/blob/master/metadetect/fitting.py#L43

esheldon: I think we only want the fitters to take single observations.

There are other codes that definitely need to take an mbobs list such as the shredder code.

But for processing different observations, I had in mind that we could do this with the new api
```python
boot = Bootstrapper(...)  # takes runners
reslist = []
for mbobs in mbobs_list:
    boot.go(mbobs)
    reslist.append(boot.fitter.get_result())
```
and then have a code to pack this list of results into an array of structs.  This is pretty compact

esheldon: basically the above code would be used in that metadetect class

beckermr: This sounds great!

ngmix/priors/priors.py
```diff
@@ -59,11 +59,10 @@ class FlatPrior(PriorBase):
...
     maxval: float
         The maximum value of the allowed range.
-    rng: np.random.RandomState or None
-        An RNG to use. If None, a new RNG is made using the numpy global RNG
-        to generate a seed.
+    rng: np.random.RandomState
+        An random number generator (RNG) to use.
     """
-    def __init__(self, minval, maxval, rng=None):
+    def __init__(self, minval, maxval, rng):
```
beckermr, line 65:
```suggestion
    def __init__(self, minval, maxval, *, rng):
```

Do we want to use something like this to preserve the API?

beckermr: We may not want to. Asking a question.

esheldon: I don't think that preserves the api does it?

Also with the way I did it one can always send the args as keyword arguments, python allows that. They can even be out of order.   e.g. this works as expected

```python
def go(x, y, z):
    print(x, y, z)

go(z=0.1, y=2.0, x=100)
```
So I think it is effectively the same either way.  So I'm not sure which is preferable or even effectively different.  What do you think?

beckermr: Ah right. Well the change allows people to pass the rng without specifying it as a kwarg. That is fine and you right that it is backwards compat.

beckermr:
@esheldon I am going to need a few days to go through this. Please do not merge.

beckermr (changes requested):
I need more time to go through this.

esheldon:
no problem

.github/workflows/mdet.yml
```diff
@@ -52,7 +52,8 @@ jobs:
       - name: test
         shell: bash -l {0}
         run: |
-          pip install git+https://github.com/esheldon/ngmix.git@master
+          # testing on ngmix2
+          pip install git+https://github.com/esheldon/ngmix.git@api-proposal
           pushd mdet_tests
```
beckermr, line 57: These lines and the ones below are why we need to shim the APIs. I forgot about this. I am doing a regression test against v1.3.8.

esheldon: I'm just doing this temporarily so I don't  get emails saying the tests failed

beckermr: This test won't run at all unless the APIs are shimmed in mdet.

esheldon: I misunderstood what was going on there; looks like you compare to a file produced with certain versions, whereas I thought you had been comparing to a fixed file

beckermr: Yep. We didn't want the file in the repo. Then we merged it by accident anyway due to my typo. 🤦

beckermr: Once 2.0.0 is released and has passed against 1.3.8, we can drop support in mdet for 1.3.8.

.github/workflows/mdet.yml
```diff
@@ -41,20 +41,17 @@ jobs:
             pytest-cov \
             meds \
             fitsio \
-            sep \
-            des-sxdes
```
beckermr: you don't need this one anymore?

mdet_tests/test_mdet_regression.py
```diff
@@ -230,12 +230,19 @@ def test_mdet_regression(fname, write=False):
     else:
         old_data = fitsio.read(fname)
         for col in old_data.dtype.names:
+            if ('psf_' in col or 'ratio' in col) and '1.3.8' in fname:
+                rtol = 1.0e-4
+                atol = 1.0e-4
+            else:
+                rtol = 1.0e-5
+                atol = 2e-6
+
```
beckermr, line 239: We should make this change in precision conditional on whether or not we are looking at the code on master vs ngmix v1.

ngmix/admom.py
```diff
@@ -32,54 +72,51 @@ class Admom(object):
...
+            A guess for the fitter.  Can be a full gaussian mixture or a single
+            value for T, in which case the rest of the parameters for the
+            gaussian are generated.
         """
         from .admom_nb import admom
 
-        guess_gmix = self._get_guess(guess)
+        if not isinstance(obs, Observation):
+            raise ValueError("input obs must be an Observation")
+        self._obs = obs
```
beckermr, line 109: Should we be caching this? Maybe the right thing here are FitResult objects that can do things like make the mixture and get the parameters etc. This way the fitters are not caching info about the data they fit. 

I am not 100% sure I like that design, but it feels odd to have the fitter not have the obs in the constructor but then cache it later.

ngmix/admom.py
```diff
@@ -112,12 +148,33 @@ def get_gmix(self):
 
         return GMixModel(pars, "gauss")
 
-    def _get_guess(self, guess):
+    def make_image(self):
```
beckermr, line 151: This could be better attached to a fit result object.

ngmix/metacal_bootstrap.py
```diff
@@ -0,0 +1,118 @@
...
+    Parameters
+    ----------
+    runner: fit runner for object
+        Must have go(obs=obs) method
+    psf_runner: fit runner for psfs
+        Must have go(obs=obs, set_result=) method
+    ignore_failed_psf: bool, optional
+        If set to True, remove observations where the psf fit fails, and
+        only fit the remaining.  Default True.
+    **metacal_kws:  keywords
```
beckermr, line 20: The rng is not in the doc string.

ngmix/metacal_bootstrap.py
```diff
@@ -0,0 +1,118 @@
...
+    obs: ngmix Observation(s)
+        Observation, ObsList, or MultiBandObsList
+    runner: ngmix Runner
+        Must have go(obs=obs) method
+    psf_runner: ngmix PSFRunner, optional
+        Must have go(obs=obs, set_result=) method
+    ignore_failed_psf: bool, optional
+        If set to True, remove observations where the psf fit fails, and
+        only fit the remaining.  Default True.
+    **metacal_kws:  keywords
```
beckermr, line 98: No RNG in the doc string.

ngmix/runners.py
```diff
@@ -0,0 +1,208 @@
...
+            fitter.go(obs=obs, guess=guess)
+        else:
+            fitter.go(obs=obs)
+
+        res = fitter.get_result()
+        if res['flags'] == 0:
+            break
+
+
+def run_psf_fitter(obs, fitter, guesser=None, ntry=1, set_result=False):
```
beckermr, line 145: Is set_result=False the right default here? It seems like unless that is true this function doesn't do a lot. What is the use case for not setting a result?

ngmix/tests/test_bootstrap.py
```diff
@@ -3,117 +3,135 @@
...
+from ngmix.fitting import LMCoellip
+from ngmix.em import GMixEM
+from ngmix.fitting import LM
+from ngmix.bootstrap import bootstrap, Bootstrapper
+from ._sims import get_model_obs
+
+FRAC_TOL = 5.0e-4
+
+
+def get_prior(*, rng, cen, cen_width, T_range, F_range):
```
beckermr, line 18: Why is this not in the `_prior.py` module?

ngmix/tests/test_em.py
```diff
@@ -0,0 +1,208 @@
...
+            assert abs(thispars[4]/truepars[4]-1) < FRAC_TOL
+            assert abs(thispars[5]/truepars[5]-1) < FRAC_TOL
+
+    # check reconstructed image allowing for noise
+    imfit = fitter.make_image()
+    imtol = 0.001 / pixel_scale**2 + noise*5
+    assert np.all(np.abs(imfit - obs.image) < imtol)
+
+
+# @pytest.mark.parametrize('noise', [0.0, 0.05])
```
beckermr, line 133: We should remove this comment and explain why 1e-6 instead of zero.

ngmix/tests/test_em.py
```diff
@@ -0,0 +1,208 @@
...
+    pars = gm.get_full_pars()
+    counts_1 = pars[0]
+
+    gm_guess = gm.copy()
+    gm_guess._data["p"] += counts_1/10 * srandu(2, rng=rng)
+    gm_guess._data["row"] += 4 * pixel_scale * srandu(2, rng=rng)
+    gm_guess._data["col"] += 4 * pixel_scale * srandu(2, rng=rng)
+    gm_guess._data["irr"] += 0.5 * pixel_scale**2 * srandu(2, rng=rng)
+    gm_guess._data["irc"] += 0.5 * pixel_scale**2 * srandu(2, rng=rng)
+    gm_guess._data["icc"] += 0.5 * pixel_scale**2 * srandu(2, rng=rng)
```
beckermr, line 90: Private data access like this is potentially very brittle. Is there a better way to do this?

ngmix/tests/test_em.py
```diff
@@ -0,0 +1,208 @@
...
+    pars = gm.get_full_pars()
+    counts = pars[0]
+
+    gm_guess = gm.copy()
+    gm_guess._data["p"] += counts/10 * srandu(rng=rng)
+    gm_guess._data["row"] += 4 * pixel_scale * srandu(rng=rng)
+    gm_guess._data["col"] += 4 * pixel_scale * srandu(rng=rng)
+    gm_guess._data["irr"] += 0.5 * pixel_scale**2 * srandu(rng=rng)
+    gm_guess._data["irc"] += 0.5 * pixel_scale**2 * srandu(rng=rng)
+    gm_guess._data["icc"] += 0.5 * pixel_scale**2 * srandu(rng=rng)
```
beckermr, line 39: ditto here from below

ngmix/tests/test_em.py
```diff
@@ -0,0 +1,208 @@
...
+    pars = gm.get_full_pars()
+    counts_1 = pars[0]
+
+    gm_guess = gm.copy()
+    gm_guess._data["p"] += counts_1/10 * srandu(2, rng=rng)
+    gm_guess._data["row"] += pixel_scale * srandu(2, rng=rng)
+    gm_guess._data["col"] += pixel_scale * srandu(2, rng=rng)
+    gm_guess._data["irr"] += 0.1 * pixel_scale**2 * srandu(2, rng=rng)
+    gm_guess._data["irc"] += 0.1 * pixel_scale**2 * srandu(2, rng=rng)
+    gm_guess._data["icc"] += 0.1 * pixel_scale**2 * srandu(2, rng=rng)
```
beckermr, line 167: ditto from above

ngmix/tests/test_em.py
```diff
@@ -0,0 +1,208 @@
...
+
+    f1 = pars[0]
+    f2 = pars[6]
+    if f1 > f2:
+        indices = [1, 0]
+    else:
+        indices = [0, 1]
+
+    # only check pars for no noise
+    # if noise == 0.0:
```
beckermr, line 185: remove comment here

proposals/fitapy_proposal.py
```diff
@@ -0,0 +1,458 @@
+"""
```
beckermr, line 1: We should remove this file.

ngmix/tests/test_metacal.py
```diff
@@ -56,15 +56,17 @@ def _get_obs(rng, set_noise_image=False):
...
     rng = np.random.RandomState(seed=100)
 
     obs = _get_obs(rng)
 
-    mpars = {
-        'psf': 'fitgauss',
-    }
-    ngmix.metacal.get_all_metacal(obs, **mpars)
+    if psf == 'galsim_obj':
+        psf = galsim.Gaussian(fwhm=0.9)
```
beckermr, line 66:
```suggestion
```

ngmix/tests/test_ml_fitting_exp_obj_gauss_psf.py
```diff
@@ -3,18 +3,26 @@
...
+# @pytest.mark.parametrize('wcs_g1', [-0.5, 0, 0.2])
+# @pytest.mark.parametrize('wcs_g2', [-0.2, 0, 0.5])
+# @pytest.mark.parametrize('g1_true', [-0.1, 0, 0.2])
+# @pytest.mark.parametrize('g2_true', [-0.2, 0, 0.1])
+@pytest.mark.parametrize('wcs_g1', [-0.5, 0])
+@pytest.mark.parametrize('wcs_g2', [-0.2, 0])
+@pytest.mark.parametrize('g1_true', [-0.1, 0])
+@pytest.mark.parametrize('g2_true', [-0.2, 0])
+@pytest.mark.parametrize('fit_model', ['exp', 'bdf', 'bd'])
+# @pytest.mark.parametrize('fit_model', ['bd'])
```
beckermr, line 22: remove this comment.

ngmix/tests/test_ml_fitting_exp_obj_gauss_psf.py
```diff
@@ -3,18 +3,26 @@
...
-@pytest.mark.parametrize('wcs_g2', [-0.2, 0, 0.5])
-@pytest.mark.parametrize('g1_true', [-0.1, 0, 0.2])
-@pytest.mark.parametrize('g2_true', [-0.2, 0, 0.1])
+from ._priors import get_prior
+
+
+# @pytest.mark.parametrize('wcs_g1', [-0.5, 0, 0.2])
+# @pytest.mark.parametrize('wcs_g2', [-0.2, 0, 0.5])
+# @pytest.mark.parametrize('g1_true', [-0.1, 0, 0.2])
+# @pytest.mark.parametrize('g2_true', [-0.2, 0, 0.1])
```
beckermr, line 16: remove these comments.

ngmix/tests/test_ml_fitting_exp_obj_gauss_psf.py
```diff
@@ -109,7 +120,12 @@ def test_ml_fitting_exp_obj_gauss_psf_smoke(
 
     g1 = np.mean(g1arr)
     g2 = np.mean(g2arr)
-    gtol = 1e-5
+
+    if fit_model == 'bd':
+        gtol = 0.002
```
beckermr, line 125: Let's add a comment on why the tolerance is worse here.

ngmix/tests/test_metacal_bootstrap.py
```diff
@@ -0,0 +1,155 @@
...
+    for key in ['noshear', '1p', '1m', '2p', '2m']:
+        assert key in resdict
+        assert resdict[key]['flags'] == 0
+
+
+@pytest.mark.parametrize('noise', [1.0e-8, 0.01])
+@pytest.mark.parametrize('use_bootstrapper', [False, True])
+def test_metacal_bootstrap_gaussmom_smoke(noise, use_bootstrapper):
+    """
+    Smoke test a Runner running the LM fitter
```
beckermr, line 118: doc string is wrong here

ngmix/tests/test_metacal_bootstrap.py
```diff
@@ -0,0 +1,155 @@
...
+    data = get_model_obs(
+        rng=rng,
+        model='gauss',
+        noise=noise,
+    )
+    obs = data['obs']
+
+    fwhm = 1.2
+    fitter = GaussMom(fwhm=fwhm)
+    psf_fitter = GaussMom(fwhm=fwhm)
```
beckermr, line 132: This is an odd choice I think.

ngmix/tests/test_metacal_bootstrap.py
```diff
@@ -0,0 +1,155 @@
...
+    else:
+        resdict = metacal_bootstrap(
+            obs=obs, runner=runner, psf_runner=psf_runner,
+            rng=rng,
+            **mcal_kws,
+        )
+
+    for key in ['noshear', '1p', '1m', '2p', '2m']:
+        assert key in resdict
+        assert resdict[key]['flags'] == 0
```
beckermr, line 155: We should add something more quantitative here I think.

ngmix/tests/test_metacal_bootstrap.py
```diff
@@ -0,0 +1,155 @@
...
```
beckermr, line 111: We should add something more quantitative here I think.

ngmix/leastsqbound.py
```diff
@@ -6,12 +6,176 @@
...
+LOGGER = logging.getLogger(__name__)
+
+
+def run_leastsq(func, guess, n_prior_pars, **keys):
+    """
+    run leastsq from scipy.optimize.  Deal with certain
+    types of errors
+
+    TODO make this do all the checking and fill in cov etc.  return
+    a dict
```
beckermr, line 39: I think this todo is done? We can remove this comment?

ngmix/leastsqbound.py
```diff
@@ -6,12 +6,176 @@
...
+
+        flags = 0
+        if ier > 4:
+            flags = 2 ** (ier - 5)
+            pars, pcov, perr = _get_def_stuff(npars)
+            LOGGER.debug(errmsg)
+
+        elif pcov0 is None:
+            # why on earth is this not in the flags?
+            flags += LM_SINGULAR_MATRIX
```
beckermr, line 82: `+=` or `|=`?

ngmix/leastsqbound.py
```diff
@@ -6,12 +6,176 @@
...
+            if dof == 0:
+                junk, pcov, perr = _get_def_stuff(npars)
+                flags |= ZERO_DOF
+            else:
+                s_sq = (fdiff[n_prior_pars:] ** 2).sum() / dof
+                pcov = pcov0 * s_sq
+
+                cflags = _test_cov(pcov)
+                if cflags != 0:
+                    flags += cflags
```
beckermr, line 108: `+=` or `|=`?

ngmix/leastsqbound.py
```diff
@@ -6,12 +6,176 @@
...
+    return pars, cov, err
+
+
+def _test_cov(pcov):
+    flags = 0
+    try:
+        e, v = np.linalg.eig(pcov)
+        (weig,) = np.where(e < 0)
+        if weig.size > 0:
+            flags += LM_NEG_COV_EIG
```
beckermr, line 169: `+=` or `|=`?

ngmix/leastsqbound.py
```diff
@@ -6,12 +6,176 @@
...
+
+        (wneg,) = np.where(np.diag(pcov) < 0)
+        if wneg.size > 0:
+            flags += LM_NEG_COV_DIAG
```
beckermr, line 173: `+=` or `|=`?

ngmix/priors/priors.py
```diff
@@ -53,17 +51,16 @@ class FlatPrior(PriorBase):
...
         PriorBase.__init__(self, rng=rng)
```
beckermr, line 64: `super` here?

I know I should have done this. Might as well do it now.

ngmix/priors/priors.py
```diff
@@ -158,12 +170,11 @@ class TwoSidedErf(PriorBase):
...
-    rng: np.random.RandomState or None
-        An RNG to use. If None, a new RNG is made using the numpy global RNG
-        to generate a seed.
+    rng: np.random.RandomState
+        An random number generator (RNG) to use.
     """
-    def __init__(self, minval, width_at_min, maxval, width_at_max, rng=None):
-        PriorBase.__init__(self, rng=rng)
+    def __init__(self, minval, width_at_min, maxval, width_at_max, rng):
+        PriorBase.__init__(self, rng)
```
beckermr, line 177: `super` here?

ngmix/tests/test_guessers.py
```diff
@@ -0,0 +1,164 @@
+import pytest
```
beckermr, line 1: We should improve the tests here. In particular, making sure the guessers are reproducible with the same RNG seed and that they produce different guesses each time is important.

ngmix/tests/test_runners.py
```diff
@@ -0,0 +1,179 @@
+import pytest
```
beckermr, line 1: I have the same comment here as I did on the guessers. We should add tests to make sure things are reproducible and change when we change the seeds.

ngmix/metacal.py
```diff
@@ -523,19 +596,12 @@ def _get_dilated_psf(self, shear, doshear=False):
 
         psf_grown_nopix = self._do_dilate(self.psf_int_nopix, shear)
 
-        if doshear and not self.shear_pixelized_psf:
+        if doshear:
             psf_grown_nopix = psf_grown_nopix.shear(g1=shear.g1,
                                                     g2=shear.g2)
 
-        if self.prepix:
```
beckermr, line 530: did the removal of this prepix option make it into the changelog?

ngmix/joint_prior.py
```diff
@@ -860,1514 +972,3 @@ def fill_fdiff(self, pars, fdiff, **keys):
         fdiff[0:index] = sqrt(chi2)
 
         return index
-
```
beckermr, line 863: Are there any class here that we need to document as removed in the changelog?

ngmix/bootstrap.py
```diff
@@ -6,2930 +6,164 @@
     - make it possible to specify the guess type (not just psf)
 
 """
-from pprint import pprint
```
beckermr, line 9: We should add all of the removed classes here to the changelog if we have not.

ngmix/tests/test_bootstrap.py
```diff
@@ -3,117 +3,135 @@
...
-        psf_im,
-        weight=psf_wt,
-        jacobian=pj,
+from ngmix import priors, joint_prior
+from ngmix.runners import Runner, PSFRunner
+from ngmix.guessers import GMixPSFGuesser, TFluxGuesser, CoellipPSFGuesser
+from ngmix.fitting import LMCoellip
+from ngmix.em import GMixEM
+from ngmix.fitting import LM
+from ngmix.bootstrap import bootstrap, Bootstrapper
```
beckermr, line 12: We need a test for the code that removes obs with bad PSF fits.

ngmix/runners.py
esheldon: Originally I had made all these functions pure by default, and returning the results instead of setting them.  But I realized that was impractical for almost all use cases we care about, because we need the gmix set.

We should make this the only behavior rather than optional.

ngmix/tests/test_em.py
esheldon: I think this was meant to be a temporary change during development, forgot to put it back

ngmix/tests/test_metacal_bootstrap.py
esheldon: what is an odd choice?

ngmix/tests/test_metacal_bootstrap.py
esheldon: what do you have in mind?

ngmix/joint_prior.py
esheldon: No, but there is a new one I will put in the new features

ngmix/bootstrap.py
esheldon: I removed all of them, good enough to say "removed all the old bootstrapper and runner classes; replaced with generic Bootstrapper, Runner, PSFRuner, and  MetacalBootstrapper`, and associated functions?

beckermr: SGTM!

ngmix/admom.py
esheldon: A fit result object would  need to hold references for the input observation(s) in order to make images

beckermr: Yup. However, the fitter would not hold the refs. If we want to apply a fitter to more than one obs, then I think a result object with these methods might be sensible. It could inherit from dict in order to preserve our apis.

ngmix/tests/test_metacal_bootstrap.py
beckermr: Fitting the PSF with Gaussian moments. IIUIC the PSF model cannot really match the actual PSF image this way.

ngmix/tests/test_metacal_bootstrap.py
beckermr: I am not sure. Maybe we could test that the images and noise images that come out are different than the input and from each other? That is not a wonderful test, but we want to make sure the output data isn't obviously wrong.

ngmix/tests/test_metacal_bootstrap.py
esheldon: This isn't used anywhere for doing metacal or getting responses, this is just a choice for a psf measure for the end user.  This is exactly what we do for metadetect when we use moments.  Are you thinking this was used internall for getting the reconvolved psf?

beckermr: Ahhhhh yes. I was thinking this was for the reconvolved PSF. Resolving!

ngmix/tests/test_metacal_bootstrap.py
esheldon: note the output observation does not have a noise image attached

beckermr: Ah right. We can still test the output image itself.

esheldon: I added image tests as well as a test of the recovered R value

beckermr: This is great!

ngmix/leastsqbound.py
```diff
@@ -6,12 +6,178 @@
...
+
+        pars, pcov0, infodict, errmsg, ier = lm_tup
+
+        if ier == 0:
+            # wrong args, this is a bug
+            raise ValueError(errmsg)
+
+        flags = 0
+        if ier > 4:
+            flags = 2 ** (ier - 5)
```
beckermr, line 78:
```suggestion
            flags |= 2 ** (ier - 5)
```

ngmix/runners.py
```diff
@@ -0,0 +1,200 @@
...
+        An object to perform measurements, must have a go(obs=obs, guess=guess)
+        method.
+    guesser: ngmix guesser object, optional
+        Must be a callable returning an array of parameters
+    ntry: int, optional
+        Number of times to try if there is failure
+
+    Side Effects
+    ------------
+    .meta['result'] is set to the fit result and the .gmix attribuite is set
```
beckermr, line 163:
```suggestion
    obs.meta['result'] is set to the fit result and the .gmix attribuite is set
```

ngmix/tests/test_bootstrap.py
```diff
@@ -4,116 +4,181 @@
...
-    boot.fit_max(
-        obj_model,
-        pars,
+    # prior = get_prior(
+    #     rng=rng,
+    #     cen=[0.0, 0.0],
+    #     cen_width=obs.jacobian.scale,
+    #     T_range=[-1.0, 1.e3],
+    #     F_range=[0.01, 1000.0],
+    # )
```
beckermr, line 76: Do you need these comments?

ngmix/tests/test_em.py
```diff
@@ -0,0 +1,198 @@
...
+import numpy as np
+
+from ngmix.em import fit_em
+from ._sims import get_ngauss_obs
+
+FRAC_TOL = 0.001
+
+
+def randomize_gmix(rng, gmix, pixel_scale):
+    gm_data = gmix.get_data()
```
beckermr, line 11: I assume this returns a reference and not a copy?

ngmix/leastsqbound.py
esheldon: in this case it doesn't matter but its good practice to always use `|=`

ngmix/tests/test_em.py
esheldon: We have not make the gmix objects non-writeable like we did for observations. Its probably not a bad idea

mdet_tests/test_mdet_regression.py
esheldon: v1 will not change right?  We already know it passes?

ngmix/admom.py
esheldon: I'm going to punt on this

ngmix/tests/test_em.py
esheldon: I opened an issue #133

mdet_tests/test_mdet_regression.py
beckermr: We test the current code in the head of the PR against master and against v1.3.8. We should demand a better tolerance for the test against master than for that against v1.3.8.

esheldon: How can one test they are in master?  Or do you mean test if version >= 2?

esheldon: oh, sorry I misunderstood, you mean if 'master' is in the filename

beckermr: yeah - I checkout master and produce a benchmark. I also install 1.3.8 and produce a benchmark. Then I compare those benchmarks to what the code in the PR produces.

esheldon: it already only uses the low tolerance for '1.3.8' in the filename

esheldon: maybe I misunderstand?

beckermr: This was my mistake. Resolving and looks good!

beckermr (approved):

beckermr:
The test failure in mdet is because master is at v1 still. It will pass when we merge.

beckermr:
Ack I want to go through a few more files.

beckermr (changes requested):
still looking sorry!

README.md
```diff
@@ -3,9 +3,8 @@ ngmix
 
 [![Build Status](https://travis-ci.com/esheldon/ngmix.svg?branch=master)](https://travis-ci.com/esheldon/ngmix)
 
-Gaussian mixture models and other code for working with for 2d images,
-implemented in python.   The code is made fast using the numba package.  Note
-the old c-extension based code is still available in the tag v0.9.5
+Gaussian mixture models and other tools for working with 2d images, implemented
+in python.   The code is made fast using the numba package.
```
beckermr, line 7:
```suggestion
in python. The code is made fast using the numba package.
```

ngmix/em.py
```diff
@@ -1,222 +1,1549 @@
...
+    ('gi', 'f8'),
+
+    # sums over all pixels
+    ('pnew', 'f8'),
 ]
-_em_conf_dtype = numpy.dtype(_em_conf_dtype, align=True)
+_sums_dtype_fluxonly = np.dtype(_sums_dtype_fluxonly, align=True)
 
 
 def test_1gauss(
```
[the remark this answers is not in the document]
esheldon: I had kept them because the provide visualization of the results, which the unit tests do not.

But maybe we an move these under and `examples/` directory along with some other examples

ngmix/guessers.py
```diff
@@ -268,49 +420,64 @@ def __call__(self, n=1, **keys):
...
         if n == 1:
             guess = guess[0, :]
         return guess
 
 
-class ParsGuesser(GuesserBase):
+class ParsGuesser(object):
     """
-    pars include g1,g2
+    Make BD guesses from the input T, flux and prior
```
beckermr, line 439: Doc string is wrong here.

ngmix/guessers.py
```diff
@@ -639,8 +705,533 @@ def __call__(self, n=1, **keys):
...
+            _fix_guess(guess, self.prior)
 
         if n == 1:
             guess = guess[0, :]
         return guess
+
+
+class GMixPSFGuesser(object):
+    """
+    guesser a full gaussian mixture for a psf.  Useful for EM and admom
```
beckermr, line 717: typos here.

ngmix/guessers.py
```diff
@@ -639,8 +705,533 @@ def __call__(self, n=1, **keys):
...
+    Parameters
+    ----------
+    rng: numpy.random.RandomState
+        Random state for generating guesses
+    ngauss: int
+        number of gaussians
+    guess_from_moms: bool, optional
+        If set to True, use weighted moments to generate the starting flux and
+        T for the guess.  If set to False, the starting flux is gotten from
+        summing the image and the fwhm of the guess isset to 3.5 times the
```
beckermr, line 728: typos isset

beckermr (approved):

PR #129, opened 2021-01-26
EM rework

### LINKED ISSUE #96: make the em code work in correct units
The original implementation works such that the fluxes of each gaussian are normalized relative to the total flux in the image.  This is inconvenient, update it to work in the correct units

I don't think this is really a breaking change because, due to the odd units, anyone using this code would eventually reset the flux anyway.

esheldon:
closes #96 

Also adds improvements to the em code from what we've learned from the shredder

TODO:

- [x] make the em runners work with new interface
- [x] add more tests
- [x] add test of em with runners
- [x] make sure bootstrappers also work
- [x] make sure docs are up to date
- [x] reconsider default maxiter and tol


esheldon:
the classes I copied from the shredder had pretty loose tolerances, which I think made sense for that scenario where you could be fitting 100 objects at once.  But maybe the defaults here should be tighter.  Certainly in the PSF "turb" test I needed tighter tolerance

ngmix/bootstrap.py
```diff
@@ -1955,44 +1955,90 @@ def go(self, ntry=1):
...
+
+    Parameters
+    ----------
+    obs: ngmix.Observation
+        The observation to fit
+    Tguess: float
+        The guess for the overall T
+    ngauss: int
+        Number of gaussians, 1 to 5
+    em_pars: dict or array
```
beckermr, line 1974: only dict, right?

ngmix/bootstrap.py
```diff
@@ -1955,44 +1955,90 @@ def go(self, ntry=1):
...
+        Parameters
+        ----------
+        ntry: int
+            Number of times to try the fit, Default 1
+        guess: ngmix.GMix
+            Use the input mixture for the first guess
+
+        Returns
+        -------
+        None
```
beckermr, line 2031:
```suggestion
```

I'd be ok with no returns here in the doc string.

ngmix/tests/test_em.py
```diff
@@ -0,0 +1,326 @@
...
+    Tpsf = 0.27
+    psf_gm = GMixModel([0.0, 0.0, 0.0, 0.0, Tpsf, 1.0], "turb")
+    # psf_gm = GMixModel([0.0, 0.0, 0.0, 0.0, Tpsf, 1.0], "gauss")
+    psf_im = psf_gm.make_image(dims, jacobian=jacob)
+
+    return Observation(psf_im, jacobian=jacob), psf_gm
+
+
+@pytest.mark.parametrize('noise', [0.0, 0.05])
+def test_1gauss(noise):
```
beckermr, line 119:
```suggestion
def test_em_1gauss(noise):
```

I've been trying to use this convention so it is easy to link the unit test functions back to the modules they target.

ngmix/tests/test_em.py
```diff
@@ -0,0 +1,326 @@
...
+    assert res['flags'] == 0
+
+    fitpars = fit_gm.get_full_pars()
+
+    if noise == 0.0:
+        tol = 1.0e-4
+        assert (fitpars[0]/pars[0]-1) < tol
+        assert (fitpars[3]/pars[3]-1) < tol
+        assert (fitpars[4]/pars[4]-1) < tol
+        assert (fitpars[5]/pars[5]-1) < tol
```
beckermr, line 157: What about the other parameters?

ngmix/tests/test_em.py
```diff
@@ -0,0 +1,326 @@
...
+
+    # check reconstructed image allowing for noise
+    imfit = fitter.make_image()
+    imtol = 0.001 / pixel_scale**2 + noise*5
+    assert np.all(np.abs(imfit - obs.image) < imtol)
+
+
+@pytest.mark.parametrize('noise', [0.0, 0.05])
+def test_2gauss(noise):
```
beckermr, line 166:
```suggestion
def test_em_2gauss(noise):
```

ngmix/tests/test_em.py
```diff
@@ -0,0 +1,326 @@
...
+
+            fitstart = indices[i]*6
+            fitend = (indices[i]+1)*6
+            thispars = fitpars[fitstart:fitend]
+
+            tol = 1.0e-4
+            assert (thispars[0]/truepars[0]-1) < tol
+            assert (thispars[3]/truepars[3]-1) < tol
+            assert (thispars[4]/truepars[4]-1) < tol
+            assert (thispars[5]/truepars[5]-1) < tol
```
beckermr, line 223: again the other pars at index 1 and 2

ngmix/tests/test_em.py
```diff
@@ -0,0 +1,326 @@
...
+
+    # check reconstructed image allowing for noise
+    imfit = fitter.make_image()
+    imtol = 0.001 / pixel_scale**2 + noise*5
+    assert np.all(np.abs(imfit - obs.image) < imtol)
+
+
+@pytest.mark.parametrize('noise', [0.0, 0.05])
+def test_2gauss_withpsf(noise):
```
beckermr, line 232:
```suggestion
def test_em_2gauss_withpsf(noise):
```

ngmix/tests/test_em.py
```diff
@@ -0,0 +1,326 @@
...
+
+            fitstart = indices[i]*6
+            fitend = (indices[i]+1)*6
+            thispars = fitpars[fitstart:fitend]
+
+            tol = 1.0e-4
+            assert (thispars[0]/truepars[0]-1) < tol
+            assert (thispars[3]/truepars[3]-1) < tol
+            assert (thispars[4]/truepars[4]-1) < tol
+            assert (thispars[5]/truepars[5]-1) < tol
```
beckermr, line 290: what about pars at index 1, 2?

beckermr (commented):
I still need to go through the EM code itself.

ngmix/tests/test_em.py
esheldon: I didn't have a guess at how far we expect those to drift.

esheldon: We can use something loose, maybe 1/10 pixel

I think the only way the center could be off but the  moments be OK is if there was a bug that changed the definition of the center

esheldon:
this PR is now superseded by #127

beckermr:
great. closing.

PR #130, opened 2021-02-18
ENH first pass at mdet regression test

beckermr:
This PR adds a regression test to the test suite. It runs mdet on a simple sim and makes sure the catalogs come out the same.


mdet_tests/test_mdet_regression.py
```diff
@@ -0,0 +1,240 @@
...
+        )
+        fnames = glob.glob(pth)
+        assert len(fnames) > 0
+        for fname in fnames:
+            old_data = fitsio.read(fname)
+            for col in old_data.dtype.names:
+                if np.issubdtype(old_data[col].dtype, np.number):
+                    assert np.allclose(
+                        all_res[col], old_data[col],
+                        atol=2e-8, rtol=1e-5,
```
beckermr, line 233:
```suggestion
                        atol=3e-8, rtol=1e-5,
```

mdet_tests/test_mdet_regression.py
```diff
@@ -0,0 +1,240 @@
...
+        )
+        fnames = glob.glob(pth)
+        assert len(fnames) > 0
+        for fname in fnames:
+            old_data = fitsio.read(fname)
+            for col in old_data.dtype.names:
+                if np.issubdtype(old_data[col].dtype, np.number):
+                    assert np.allclose(
+                        all_res[col], old_data[col],
+                        atol=5e-8, rtol=1e-5,
```
beckermr, line 233:
```suggestion
                        atol=1e-6, rtol=1e-5,
```

mdet_tests/test_mdet_regression.py
```diff
@@ -0,0 +1,240 @@
...
+        )
+        fnames = glob.glob(pth)
+        assert len(fnames) > 0
+        for fname in fnames:
+            old_data = fitsio.read(fname)
+            for col in old_data.dtype.names:
+                if np.issubdtype(old_data[col].dtype, np.number):
+                    assert np.allclose(
+                        all_res[col], old_data[col],
+                        atol=1e-6, rtol=1e-5,
```
beckermr, line 233:
```suggestion
                        atol=1e-5, rtol=1e-5,
```

mdet_tests/test_mdet_regression.py
```diff
@@ -0,0 +1,240 @@
...
+        )
+        fnames = glob.glob(pth)
+        assert len(fnames) > 0
+        for fname in fnames:
+            old_data = fitsio.read(fname)
+            for col in old_data.dtype.names:
+                if np.issubdtype(old_data[col].dtype, np.number):
+                    assert np.allclose(
+                        all_res[col], old_data[col],
+                        atol=1e-5, rtol=1e-5,
```
beckermr, line 233:
```suggestion
                        atol=2e-6, rtol=1e-5,
```

beckermr:
So the benchmark here is from running version 1.3.8 on osx.

If I run the same version on linux, I can get set a lower tolerance (around 3-e8). 

For what is on master now, a value at 2e-6 is needed for the test to pass. 

Thus we have already seen some changes, but they are at a level that is probably ok to ignore.

We could also change the test to benchmark against what is in master now as well. 

What do you think @esheldon?

esheldon:
should we remake the reference file on linux?

beckermr:
> should we remake the reference file on linux?

Sure. This won't change much but is probably good for reproducing things.

.github/workflows/mdet.yml
```diff
@@ -0,0 +1,61 @@
...
+          pip install git+https://github.com/esheldon/metadetect.git \
+            --force-reinstall --no-deps --no-cache-dir
+
+          python -m pip install -e .
+
+      - name: test
+        shell: bash -l {0}
+        run: |
+          pip install git+https://github.com/esheldon/ngmix.git@master
+          cd mdet_tests
```
beckermr, line 55:
```suggestion
          pushd mdet_tests
```

beckermr (commented):
One but

beckermr:
OK. This is ready to go. The tests check against 1.3.8 and the current code on master.

esheldon:
I think the test itself looks good.

Why is there also a notebook?

My main complaint is that there are now multiple versions of this rather large file in the history.

beckermr:
We can squash down the history. The notebook was for me to make a quick plot of the simulated image to make sure it works.

beckermr:
Let me change the test to build the benchmark on the fly.

beckermr:
OK ready for another look. Do not merge with squshing.

beckermr:
I removed the big files in the repo here. We ok to merge?

PR #132, opened 2021-02-23
REF use logging

beckermr:
This PR starts in on using logging in the fitting module.

### DIFF
--- a/ngmix/fitting.py
+++ b/ngmix/fitting.py
@@ -9,6 +9,7 @@ from numpy import exp, sqrt, where, isfinite
 from numpy import linalg
 from numpy.linalg import LinAlgError
 from pprint import pformat
+import logging
 
 from . import gmix
 from .gmix import GMixList, MultiBandGMixList
@@ -24,6 +25,8 @@ from . import stats
 
 from .gmix_nb import gmix_convolve_fill
 from .fitting_nb import fill_fdiff
+
+LOGGER = logging.getLogger(__name__)
 
 MAX_TAU = 0.1
 MIN_ARATE = 0.2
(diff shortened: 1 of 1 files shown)


ngmix/fitting.py
```diff
@@ -3401,21 +3404,25 @@ def _set_minmax_err(self, min_err, max_err):
...
-        stream.write("%s\n" % None)
+        txt += "%s\n" % None
     else:
-        # fmt = ' '.join( [fmt+' ']*len(pars) )
-        # stream.write(fmt % tuple(pars))
         s = format_pars(pars, fmt=fmt)
-        stream.write(s)
-        stream.write("\n")
+        txt += s
+        txt += "\n"
```
esheldon, line 3420: do we need the newline with the logger?

beckermr: ah probably not

ngmix/fitting.py
```diff
@@ -3401,21 +3404,25 @@ def _set_minmax_err(self, min_err, max_err):
...
     """
+    txt = ""
     if front is not None:
-        stream.write(front)
-        stream.write(" ")
+        txt += front
+        txt += " "
     if pars is None:
-        stream.write("%s\n" % None)
+        txt += "%s\n" % None
```
beckermr, line 3416:
```suggestion
        txt += "%s" % None
```

ngmix/fitting.py
beckermr, line 3420:
```suggestion
```

esheldon:
how do we control the log level for these loggers that are created in a single module?  Do we have to set it on that logger itself or is there a way for an application to control the level of all the loggers?

beckermr:
> how do we control the log level for these loggers that are created in a single module? Do we have to set it on that logger itself or is there a way for an application to control the level of all the loggers?

All loggers are globals. So to control the level, you get the logger for the module and set the level. 

So things like

```
import logging

logger = logging.getLogger("ngmix.fitting")
# do thing with logger
```

esheldon:
So there is no way to turn on debug level logging everywhere?

beckermr:
Yes. You set it for the ngmix logger from `logging.getLogger("ngmix")`

beckermr:
Loggers also inherit settings from parents.

esheldon:
ok, nice.

PR #137, opened 2021-03-10
metacal default psf set to gauss

esheldon:


beckermr (approved):
LGTM! Not sure if you are done with it yet or not ofc.

PR #143, opened 2021-03-12
fitters/runners/bootstrappers return FitModel/Result objects

esheldon:
This was suggested by @beckermr 

TODO

- [x] make the psf runners/bootstrappers return results/lists of results.  Only set result if requested (old behavior)
- [x] docs
- [x] update examples
- [x] port over metadetect tests

### DIFF
--- a/ngmix/metacal/bootstrap.py
+++ b/ngmix/metacal/bootstrap.py
@@ -43,7 +43,7 @@ class MetacalBootstrapper(object):
         obs: ngmix Observation(s)
             Observation, ObsList, or MultiBandObsList
         """
-        self._resdict, self._obsdict = metacal_bootstrap(
+        return metacal_bootstrap(
             obs=obs,
             runner=self.runner,
             psf_runner=self.psf_runner,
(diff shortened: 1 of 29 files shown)


ngmix/metacal/bootstrap.py
```diff
@@ -142,10 +110,10 @@ def metacal_bootstrap(
     resdict = {}
 
     for key, tobs in obsdict.items():
-        bootstrap(
+        resdict[key] = bootstrap(
             obs=tobs, runner=runner, psf_runner=psf_runner,
             ignore_failed_psf=ignore_failed_psf,
         )
-        resdict[key] = runner.get_result()
+        # resdict[key] = runner.get_result()
```
beckermr, line 117: remove this comment

beckermr (changes requested):
My main comment is that if we are going to do this, we need a FitResult base class that every fitter uses. Probably a similar one for the Galsim stuff.

esheldon:
There are no required methods.  For example the gaussmom result is a pure dictionary.  Also the teamplate fitters return `FitModel` that, while they have methods that are used during fitting, are not useful for the end user.

So the base class would literally by
```python
class FitResult(dict):
    pass
```
For this reason I decided not to use a base class.  But we can if you think it is better.

beckermr:
Hmmmm ok. Let me look again.

beckermr:
I see your point on little to few shared methods. The only last bit to consider is if we'd ever want to right code that generically processes the results of fits, e.g.

```
if isinstance(blah, ngmix.FitResult):
    # do something
```

Right now, this won't be possible because the generic base class is absent. 

We could use a generic base class even if it only identifies the kind of thing each object is.

esheldon:
Yeah, I know what you mean.  I used to use that pattern a lot.

Generally pythonistas don't recommend using isinstance any more because of "easier to ask for forgiveness than permission"

In that vein, we advertise that we return a dict-like, and if they want to test for some feature in the returned value they do

```python
res = fitter.go(obs)
try:
    gm = res.get_gmix()
    # do something
except:
    # do something else
```

beckermr:
OK shrug. Whatever makes you happiest! :)

beckermr (approved):

esheldon:
I don't really care either way, just following advice of experts

esheldon:
If you feel strongly, go ahead and modify the PR

beckermr:
Nope no worries from me at all! Merge away!

PR #144, opened 2021-03-15
add metadetect example

esheldon:


esheldon:
actions isn't running right now, we'll wait till it comes back up

PR #145, opened 2021-03-16
Flux units

### LINKED ISSUE #95: make ngmix work in flux units
I think this can be accomplished by putting in the correct pixel area factors when processing each image.

currently we are instead modifying the data so it is surface brightness (e.g. in fitvd), but might as well do this in the code.

esheldon:
closes #95 

This is a big, breaking change.

The pixel area is carried through a number of calculations now such that the units are in flux, like galsim.  However, no integration of pixels is performed.

### DIFF
--- a/ngmix/guessers.py
+++ b/ngmix/guessers.py
@@ -782,7 +782,7 @@ class GMixPSFGuesser(object):
         of the image
         """
         scale = obs.jacobian.scale
-        flux = obs.image.sum() * scale**2
+        flux = obs.image.sum()
         # for DES 0.9/0.263 = 3.42
         fwhm = scale * 3.5
         T = moments.fwhm_to_T(fwhm)
@@ -803,6 +803,10 @@ class GMixPSFGuesser(object):
         wt = GMixModel([0.0, 0.0, 0.0, 0.0, Tweight, 1.0], "gauss")
 
         res = wt.get_weighted_moments(obs=obs, maxrad=1.0e9)
+        assert res['flags'] == 0, 'weighted moments failed'
+
+        # ngmix is in flux units, need to divide by area
+        area = scale**2
 
         Tmeas = res['T']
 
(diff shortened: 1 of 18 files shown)


esheldon:
the mdet test will fail for fluxes

ngmix/guessers.py
```diff
@@ -803,6 +803,10 @@ def _get_T_flux_from_moms(self, obs):
         wt = GMixModel([0.0, 0.0, 0.0, 0.0, Tweight, 1.0], "gauss")
 
         res = wt.get_weighted_moments(obs=obs, maxrad=1.0e9)
+        assert res['flags'] == 0, 'weighted moments failed'
```
beckermr, line 806: This should raise an appropriate error.

ngmix/tests/test_ml_fitting_exp_obj_gauss_psf.py
```diff
@@ -64,6 +64,8 @@ def test_ml_fitting_exp_obj_gauss_psf_smoke(
     guess[4] = fwhm_to_T(0.5)
 
     if fit_model == 'bd':
+        guess[5] = 1.0
+        guess[6] = 0.5
```
beckermr, line 68: Was this a bug or something?

beckermr (commented):
Random comments. This is a bold PR. I like it!

ngmix/tests/test_ml_fitting_exp_obj_gauss_psf.py
esheldon: the guesses from the prior were not very good

esheldon:
for the mdet tests I think we might need to ignore the flux entry in the pars when comparing to 1.3.8

beckermr (approved):

esheldon:
I don't know if the mdet test will pass after merging to master, but can do a separate PR to address that

PR #150, opened 2021-03-18
More tests

esheldon:
TODO:

- [ ] get all major modules > 95% for stuff we care about
- [ ] tweak coverage for parts we don't care about (plotting etc.)


esheldon:
We have good test coverage other than medsreaders.py

beckermr:
Do you want me to take a pass at writing tests for those?

esheldon:
Do you have ideas?

.coveragerc
```diff
@@ -3,3 +3,10 @@ omit =
     # we omit numba code since we cannot measure it
     ngmix/*_nb.py
     ngmix/fastexp.py
+    ngmix/leastsqbound.py
```
beckermr, line 6: This module is important for like every fit in ngmix. Do we really want to ignore it?

.coveragerc
```diff
@@ -3,3 +3,10 @@ omit =
     # we omit numba code since we cannot measure it
     ngmix/*_nb.py
     ngmix/fastexp.py
```
beckermr, line 5: We should be testing this function but I guess this is a numab issue?

.github/workflows/test.yml
```diff
@@ -41,6 +41,7 @@ jobs:
             galsim \
             scikit-learn \
             emcee \
+            fitsio \
```
beckermr, line 44: Do we need to add this to the conda package deps too?

esheldon: that is for testing the save_mixture, load_mixture for the gmix_ndim stuff.

I'm not sure anyone even uses that class, so I'm not sure if its worth having it in the deps

esheldon: We will also need it for testing the medsreaders

.coveragerc
esheldon: yes, for the coverage printouts

.coveragerc
esheldon: Its hard to test most of the stuff in there.  Its all branches for various ways the fitter can behave.  I didn't know how to trigger them.

esheldon: I think coverage was only about 30%

ngmix/gmix.py
```diff
@@ -908,11 +905,11 @@ def __init__(self, pars, model):
...
     def _do_init(self, pars, model):
-        self._model = _gmix_model_dict[model]
-        self._model_name = _gmix_string_dict[self._model]
+        self._model = get_model_num(model)
+        self._model_name = get_model_name(self._model)
 
-        self._ngauss = _gmix_ngauss_dict[self._model]
-        self._npars = _gmix_npars_dict[self._model]
+        self._ngauss = get_model_ngauss(self._model)
+        self._npars = get_model_npars(self._model)
```
beckermr, line 912: How did our tests miss this?

ngmix/joint_prior.py
```diff
@@ -977,3 +988,44 @@ def fill_fdiff(self, pars, fdiff):
         fdiff[0:index] = sqrt(chi2)
 
         return index
+
+    def sample(self, n=None):
```
beckermr, line 992: The keyword `n` should be `nrand` here to match the other priors.

ngmix/simplex.py
```diff
@@ -1,198 +0,0 @@
-"""
```
beckermr, line 1: This needs a changelog entry.

ngmix/tests/test_admom.py
```diff
@@ -66,17 +68,27 @@ def test_admom_smoke(g1_true, g2_true, wcs_g1, wcs_g2):
...
 
                 g1arr.append(_g1)
                 g2arr.append(_g2)
                 Tarr.append(_T)
-        except GMixRangeError:
+
+                fim = res.make_image()
+                assert fim.shape == im.shape
+
+        except GMixRangeError:  # pragma: no cover
```
beckermr, line 83: So for tests, coverage should be complete with no exclusions or we should remove the test code. A test is worthless if we don't execute all of its parts.

ngmix/gmix.py
esheldon: that wasn't wrong.

Using the functions is better because they raise a better error.  Using the dict directly would just raise a KeyError which isn't very informative

ngmix/joint_prior.py
esheldon: We should make this uniform everywhere.  What should it be?  `n` or `nrand` or `size` ?

ngmix/tests/test_gmix.py
```diff
@@ -32,7 +33,7 @@ def test_gmix_get_pars(model, row, col, flux, g1, g2, T):
     elif model == 'turb':
         assert len(gm) == 3
     else:
-        raise AssertionError("len(gm) was not checked!")
+        raise AssertionError("len(gm) was not checked!")  # pragma: no cover
```
beckermr, line 36: same comment

ngmix/tests/test_gmix.py
```diff
@@ -60,7 +61,7 @@ def test_gmix_get_pars(model, row, col, flux, g1, g2, T):
         assert len(data) == 3
         assert len(full_pars) == 3*6
     else:
-        raise AssertionError("len(gm) was not checked!")
+        raise AssertionError("len(gm) was not checked!")  # pragma: no cover
```
beckermr, line 64: same comment

ngmix/tests/test_guessers.py
```diff
@@ -29,35 +37,53 @@ def test_noprior_guessers_smoke(guesser_type, nband):
...
             flux_center = [1.0]*nband_use
             nu_center = 1
             guesser = guessers.R50NuFluxGuesser(
                 rng=rng, r50=r50_center, nu=nu_center, flux=flux_center,
+                prior=prior,
             )
             npars = 6 + nband_use
         else:
-            raise ValueError('bad guesser %s' % guesser_type)
+            raise ValueError('bad guesser %s' % guesser_type)  # pragma: no cover
```
beckermr, line 78: same comment

ngmix/tests/test_guessers.py
```diff
@@ -130,7 +156,7 @@ def test_prior_guessers_smoke(guesser_type, nband):
...
```
beckermr, line 159: same comment

ngmix/tests/test_guessers.py
```diff
@@ -176,7 +202,7 @@ def test_psf_guessers_smoke(guesser_type, ngauss, guess_from_moms):
             )
             npars = get_coellip_npars(ngauss)
         else:
-            raise ValueError('bad guesser %s' % guesser_type)
+            raise ValueError('bad guesser %s' % guesser_type)  # pragma: no cover
```
beckermr, line 205: same comment

ngmix/tests/test_admom.py
esheldon: How do you trigger that?

ngmix/tests/test_priors_base.py
```diff
@@ -34,4 +34,5 @@ def test_priors_gpriorbase_raises():
         with pytest.raises(RuntimeError) as e:
             args = tuple([None] * num)
             mth(*args)
-            assert "over-ride me" in e.value
+            # unreachable
+            assert "over-ride me" in e.value  # pragma: no cover
```
beckermr, line 38: If we can never get here then we don't need the test, right?

ngmix/tests/test_priors_shear.py
```diff
@@ -42,14 +42,19 @@ def test_priors_gpriorgauss():
 
     with pytest.raises(NotImplementedError) as e:
         pr.sample1d()
-        assert "no 1d for gauss" in e.value
+        # unreachable
```
beckermr, line 45: same comment

ngmix/tests/test_runners.py
```diff
@@ -123,7 +123,7 @@ def test_runner_lm_simple(model, psf_model_type, noise, guesser_type):
                 T=0.25,
             )
         else:
-            raise ValueError('bad guesser')
+            raise ValueError('bad guesser')  # pragma: no cover
```
beckermr, line 126: same comment

beckermr (changes requested):
I left some comments.

ngmix/joint_prior.py
beckermr: I have been moving all functions to `nrand`.

ngmix/tests/test_admom.py
esheldon: I guess another way to proceed would be to not catch the exception at all.

It is currently never being raised which is why it wasn't covered. But on some other platform it might be raised due to some small difference in floating point operations

esheldon: Or we could take off the pragma and just be happy with lower coverage fraction

ngmix/tests/test_priors_base.py
esheldon: right, I wasn't sure why you had that there at all since it should be unreachable

ngmix/tests/test_admom.py
beckermr: I think you made a change in Admom itself in this PR that will always catch this error. Thus I think there is no reason to catch it at all in the tests.

ngmix/joint_prior.py
```diff
@@ -60,7 +60,7 @@ def set_bounds(self):
 
         self.bounds = bounds
 
-    def get_widths(self, n=10000):
+    def get_widths(self, nrand=10000):
```
beckermr, line 63: You need to change the doc string too.

ngmix/joint_prior.py
```diff
@@ -187,7 +187,7 @@ def get_lnprob_array(self, pars):
 
         return lnp
 
-    def sample(self, n=None):
+    def sample(self, nrand=None):
```
beckermr, line 190: Same comment on doc strings.

ngmix/joint_prior.py
```diff
@@ -422,7 +422,7 @@ def get_lnprob_array(self, pars):
...
```
beckermr, line 425: Same comments on docstrings.

ngmix/joint_prior.py
```diff
@@ -619,7 +619,7 @@ def get_lnprob_array(self, pars):
...
```
beckermr, line 622: docstrings

ngmix/joint_prior.py
```diff
@@ -815,7 +815,7 @@ def get_lnprob_array(self, pars):
...
```
beckermr, line 818: docstrings

ngmix/joint_prior.py
```diff
@@ -989,7 +989,7 @@ def fill_fdiff(self, pars, fdiff):
 
         return index
 
-    def sample(self, n=None):
+    def sample(self, nrand=None):
```
beckermr, line 992: docstrings

ngmix/tests/test_admom.py
esheldon: actually for this particilar test in admom, that can never be raised.  I now am catching it in the admom code.  I'll remove this

esheldon: actually that was to catch the conversion from e1, e2 to g1, g2

I think we can safely ignore it

beckermr: If we can ignore, then let's remove it.

ngmix/joint_prior.py
beckermr: Why is this reolved? The docstring still says `n` and not `nrand`.

esheldon: Yeah, it says `nrand: ` now

beckermr (approved):
Ignore my last comment. Weird github UI thing.

esheldon:
oh, it was in guessers.py not joint_priors.py

beckermr:
Ohhhhh ok.

.coveragerc
esheldon: I think I'll put it back in to remind us to divise a way to test it.

But I don't have ideas

esheldon:
admom seems to be failing suddenly for python 3.6

beckermr:
Admom failure is tiny

> assert 1.0578204982236716e-06 < 1e-06

I'd increase the tolerance and move on, but that's me.

esheldon:
It passed this time.  Maybe there is some homogeneity in the machines used by github actions

esheldon:
got leastsqbound up to 78%

ngmix/tests/test_leastsqbound.py
```diff
@@ -0,0 +1,274 @@
...
+        try:
+            res = bootstrap(obs=obs, runner=runner, psf_runner=psf_runner)
+            allflags[i] = res['flags']
+        except ngmix.BootPSFFailure:
+            allflags[i] = 1
+
+    assert np.any(allflags != 0)
+
+
+def test_leastsqbound_errors():
```
beckermr, line 178: Shouldn't the output flags be tested here or something?

esheldon: yes, just out of time for today

beckermr: ahh sorry to jump the gun! Have a good weekend!

esheldon:
this ones getting too big. gonna merge and we can add more tests in different PRs

PR #151, opened 2021-03-22
Refactoring and renaming

### LINKED ISSUE #149: rename some classes
Names like `LM` are confusing

* `LM` -> `Fitter`
* `LMCoellip` -> `CoellipFitter`
* `GMixEM` -> `EMFitter`
* `GMixEMFixCen` -> `EMFitterFixCen`
* `GMixEMFluxOnly` -> `EmFitterFluxOnly`
* `TemplateFluxFitter` -> `PSFFluxFitter` (we can still support non-psf flux fitting)

esheldon:
closes #149 

Renaming some classes.  I also decided to put related modules together into sub-packages.  E.g. we generally keep the files with numba jitted functions in a separate module.


esheldon:
the mdet test cannot pass until we merge this to master

esheldon:
Now is the time to bikeshed

esheldon:
These are stylistic changes so have at it!

ngmix/defaults.py
```diff
@@ -8,3 +8,20 @@
 # for priors etc.
 LOWVAL = -np.inf
 BIGVAL = 9999.0e47
+
+DEFAULT_LM_PARS = {"maxfev": 4000, "ftol": 1.0e-5, "xtol": 1.0e-5}
+
+ADMOM_RESULT_DTYPE = [
```
beckermr, line 14: I'm not 100% sure this default belongs here.

esheldon: This is a result of the refactoring into sub-packages.

I had to take it out of the package to use it in another package due to circular imports.

esheldon: wasn't a problem when they were plain modules, something special about packages

beckermr: ah right. Fine by me ofc. Usually, a circle import indicates a design issue, but I don't think we should care.

esheldon: could make another module `types.py` or something and put others there as well

esheldon: looking again there is  no reason to re-use this; it is I think OK to just copy it

beckermr (approved):

ngmix/defaults.py
beckermr: Ahhhh! I think that was the design flaw. They are simply not the same thing. Too DRY!

esheldon: agreed, too dry :)

esheldon:
Its important to turn on the brain when refactoring :)

PR #156, opened 2021-04-20
ENH FFT-based pre-PSF moments

beckermr:
This PR adds an FFT-based measure of pre-PSF moments.


beckermr:
@esheldon @rmjarvis Can you all provide some code review here?

ngmix/tests/test_prepsfmom.py
```diff
@@ -0,0 +1,344 @@
...
+
+    mom_cov = np.cov(np.array(momarr).T)
+    print("mom cov ratio:\n", res["mom_cov"]/mom_cov, flush=True)
+    assert np.allclose(res["mom_cov"], mom_cov, atol=0, rtol=0.5)
+    assert np.allclose(np.mean(farr), flux_true, atol=0, rtol=0.1)
+
+    # at low SNR we get a lot of problems with division by noisy things
+    if snr > 100:
+        assert np.allclose(np.mean(Tarr), T_true, atol=0, rtol=0.1)
+        assert np.allclose(np.mean(g1arr), g1_true, atol=0, rtol=0.2)
```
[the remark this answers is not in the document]
beckermr: I don't think this is crazy. If we have say 10% accuracy in the raw <xx-yy> moment and 10% accuracy in <xx+yy> then when we measure g1 which is <xx-yy>/<xx+yy>, the accuracy will get worse.

ngmix/prepsfmom/gaussmom.py
```diff
@@ -0,0 +1,352 @@
...
+            Only valid when `direct_deconv` is False. Defaults to False.
+
+        Returns
+        -------
+        result dictionary
+        """
+        if not isinstance(obs, Observation):
+            raise ValueError("input obs must be an Observation")
+
+        if not obs.has_psf():
```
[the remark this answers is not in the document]
beckermr: This is a really good point. I will remove this feature.

ngmix/prepsfmom/gaussmom.py
```diff
@@ -0,0 +1,352 @@
...
+    # finally we build the kernels, moments and their errors
+    df = f[1] - f[0]  # this is the area factor for the integral we are
+    # doing in Fourier space
+
+    # build the flux, radial, plus and cross kernels / moments
+    fkf = kernels["fkf"]
+    fkr = kernels["fkr"]
+    fkp = kernels["fkp"]
+    fkc = kernels["fkc"]
+    mf = np.sum(imfft * fkf).real * df**2
```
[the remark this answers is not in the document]
beckermr: I do apply a phase shift in fourier space to translate the kernels to the right spots for the sum. 

I don't see how this would cause extra noise though.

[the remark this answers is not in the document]
esheldon: OK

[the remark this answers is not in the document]
beckermr: I should have said, I don't understand how this would cause extra noise. I didn't mean to imply it doesn't. I don't know either way.

beckermr:
ok @esheldon @rmjarvis this is ready for another look.

esheldon (approved):
I'll approve but I eagerly wait Mike's review as he has more expertise with this type of work

ngmix/prepsfmom/gaussmom.py
```diff
@@ -0,0 +1,347 @@
...
+
+    If the fwhm of the weight function is of similar size to the PSF or smaller,
+    then the object properties returned by this fitter will be very noisy.
+
+    Parameters
+    ----------
+    fwhm : float
+        The size of the kernel. This parameter has a slightly different meaning
+        for each kernel. Roughly each parameter corresponds to the FWHM of the
+        kernel.
```
beckermr, line 26:
```suggestion
        The FWHM of the Gaussian kernel.
```

beckermr:
OK @esheldon this is ready for another look!

changes since last time:
 - I vastly simplified the code and it is significantly faster now. 
 - I consulted with Mike on the Fourier-space handling and my implementations are consistent in methodology with galsim except for one detail that Mike and I both agree doesn't matter. (I preserve the phase of the PSF amplitude when truncating before deconvolution, galsim does not.)
 - I still do not understand a detail of the covariance computation for the measurements, but the one I have appears to predict the correct error to within 10-20% in a variety of contexts and at very high precision (sub 2%) if you take one example an run a bunch of monte carlo realizations.
 - I put in a bunch of flagging and error handling consistent with how the GaussMom does stuff.

beckermr:
The mdet test failure is addressed in #158

ngmix/ksigmamom.py
```diff
@@ -0,0 +1,445 @@
...
+                raise RuntimeError(
+                    "The PSF and observation must have the same WCS "
+                    "Jacobian for measuring pre-PSF moments."
+                )
+        else:
+            psf_obs = None
+
+        return self._meas_fourier_only(obs, psf_obs, return_kernels)
+
+    def _meas_fourier_only(self, obs, psf_obs, return_kernels):
```
esheldon, line 69: what does the `_only` part of the name imply here?

ngmix/ksigmamom.py
```diff
@@ -0,0 +1,445 @@
...
+                if k == "msk":
+                    continue
+                new_kres[k] = np.zeros((fft_dim, fft_dim), dtype=np.complex128)
+                new_kres[k][kres["msk"]] = kres[k]
+            res["kernels"] = new_kres
+
+        return res
+
+
+def _measure_moments_fft(kim, kpsf_im, tot_var, eff_pad_factor, kernels, drow, dcol):
```
esheldon, line 153: I think this function could be refactored, e.g. pack the results into a dict in a separate function.  But not big deal

ngmix/ksigmamom.py
```diff
@@ -0,0 +1,445 @@
...
+    df2 = df * df
+    df4 = df2 * df2
+
+    # we only sum where the kernel is nonzero
+    fkf = kernels["fkf"]
+    fkr = kernels["fkr"]
+    fkp = kernels["fkp"]
+    fkc = kernels["fkc"]
+
+    mf = np.sum((kim * fkf).real) * df2
```
esheldon, line 189: bunch of temporary arrays here, inefficient.  Have you profiled to see where the most time is spent?  Is it the ffts?

ngmix/ksigmamom.py
```diff
@@ -0,0 +1,445 @@
...
+        #
+        #  im_cen_phase / psf_im_cen_phase = exp(ik * (im_cen - psf_cen))
+        #
+        # and then multiply it into the image.
+        #
+        # This operation and the deconvolutiomn will be done
+        # later in _measure_moments_fft
+
+        # now build the kernels
+        kres = _ksigma_kernels(
```
esheldon, line 121: maybe call this "kernels" to be consistent with the moments measuring code

ngmix/tests/test_ksigmamom.py
```diff
@@ -0,0 +1,317 @@
...
+@pytest.mark.parametrize('snr', [1e2])
+@pytest.mark.parametrize('pixel_scale', [0.25])
+@pytest.mark.parametrize('fwhm,psf_fwhm', [(2.0, 1.0)])
+@pytest.mark.parametrize('mom_fwhm', [2.0])
+@pytest.mark.parametrize('image_size', [58])
+@pytest.mark.parametrize('pad_factor', [1.5])
+def test_ksigmamom_mn_cov(
+    pad_factor, image_size, fwhm, psf_fwhm, pixel_scale, snr, mom_fwhm,
+):
+    """Slower test to make sure means and errors are right in a well-sampled case."""
```
esheldon, line 198: does this mean that the pixelization is the major reason values and errors are incorrect?

esheldon (changes requested):
Overall looks fine. I had a few comments and questions

About speed:  How fast are these per image?  Can you post a profiler run?  I'm interested where the time is being spent.

beckermr:
Here is a run through the profiler.

```
        625 function calls (621 primitive calls) in 0.254 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        2    0.142    0.071    0.142    0.071 {built-in method scipy.fft._pocketfft.pypocketfft.c2c}
        1    0.046    0.046    0.047    0.047 ksigmamom.py:350(_ksigma_kernels)
        4    0.014    0.003    0.014    0.003 arraypad.py:454(_as_pairs)
        1    0.011    0.011    0.024    0.024 ksigmamom.py:153(_measure_moments_fft)
        4    0.008    0.002    0.008    0.002 arraypad.py:129(_set_pad_area)
        1    0.008    0.008    0.008    0.008 ksigmamom.py:305(_compute_cen_phase_shift)
        2    0.006    0.003    0.006    0.003 arraypad.py:86(_pad_simple)
        1    0.005    0.005    0.247    0.247 ksigmamom.py:69(_meas_fourier_only)
        1    0.004    0.004    0.251    0.251 ksigmamom.py:32(go)
        1    0.003    0.003    0.254    0.254 <string>:1(<module>)
        1    0.002    0.002    0.002    0.002 ksigmamom.py:201(<listcomp>)
       18    0.001    0.000    0.001    0.000 {method 'reduce' of 'numpy.ufunc' objects}
        1    0.001    0.001    0.001    0.001 ksigmamom.py:334(_deconvolve_im_psf_inplace)
       10    0.000    0.000    0.000    0.000 jacobian.py:302(_finish_init)
```

This is for an image that is 315 pixels on a side w/ 4x padding. Most of the time is spent doing the FFT and then making the kernels.

ngmix/ksigmamom.py
beckermr: Good catch! This is leftover from previous code versions. I renamed it.

ngmix/ksigmamom.py
beckermr: Yes. Most of the time is spent doing FFTs or generating the kernels themselves.

ngmix/tests/test_ksigmamom.py
beckermr: "Well-sampled" refers to running a bunch of monte carlo realizations, not the pixels. I will clarify.

ngmix/ksigmamom.py
beckermr: done.

beckermr:
For a 48x48 image, the timing is

```
3.45 ms ± 703 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

beckermr:
The GaussMom code on the same obs has

```
959 µs ± 149 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

So the new code is slower, mostly due to the FFTs.

beckermr:
For sure I think the code could be refactored to do all of the kernel generation, deconvolution, and sums in a single loop in numba. This might save ~30-40% in the runtime at most, since half of the time is FFTs already. This savings would be non-trivial OFC since this bit of code will be run literally billions of times. Shall we go for it?

esheldon:
In my opinion, since the code is already pretty fast, I would only refactor to numba if it will make the code easier to read.

beckermr:
numba will make the code harder to read for me, but maybe not for others?

esheldon:
I often use it when lots of broadcasting or index tricks are required for numpy only.  Sometimes those things obscure the basic formula you are trying to implement, whereas with numba the inner loop can be a bare expression of the formula

esheldon:
If that's not the case, then leave the code as is!

beckermr:
OK let me think on it. I am willing to spend time here since I plan to use this code a lot.

esheldon (approved):

PR #161, opened 2021-06-21
BUG fix double convolution of pixel in metacal

### LINKED ISSUE #160: Possible double pixel convolution in metacal images?
Going over the different steps in the metacal procedure, I got a bit puzzled by this line where if I understand correctly the output images are drawn:
https://github.com/esheldon/ngmix/blob/8e81c5337b022e06c517c4abb8a7bdae7204d737/ngmix/metacal/metacal.py#L303

What puzzles me is when drawn here, the shear galaxy image should have already been reconvolved with the target PSF, which already contains the pixel response (added [here](https://github.com/esheldon/ngmix/blob/8e81c5337b022e06c517c4abb8a7bdae7204d737/ngmix/metacal/metacal.py#L252)), so I  would expect a `method="no_pixel"` when drawing this galaxy.

To make things clearer for me, I tried to do a pure GalSim equivalent, starting from the metacal example script [here](https://github.com/esheldon/ngmix/blob/master/examples/metacal/metacal.py). Given some `obs` and `obsdict` obtained by metacal, to reproduce the output with GalSim I need to do:

``` python
# Grab observations
im   = galsim.InterpolatedImage(galsim.Image(obs.image, scale=0.263))
psf  = galsim.InterpolatedImage(galsim.Image(obs.psf.image, scale=0.263))
rpsf = obsdict['noshear'].psf.galsim_obj 
# Deconvolve input image by PSF+pixel
gal = galsim.Convolve([im, galsim.Deconvolve(psf)])
# Apply shear and reconvolve by target PSF !!!which should contain the pixel response as well !!!
gal_1p = galsim.Convolve([gal.shear(g1=0.01,g2=0), rpsf])
# Draw the postage stamp
rec_1p = gal_1p.drawImage(nx=42,ny=42, scale=0.263).array
```
which gives me the following residuals when compared to `obsdict['1p'].image`:
image

but note that I draw the galaxy without the `method="no_pixel"` flag despite the reconvolution PSF being already pixel convolved.

Am I missing something? or is the output metacal image convolved with the pixel response twice (which doesn't matter that much probably)?

esheldon:
closes #160

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -35,6 +35,9 @@
 
 ### bug fixes
 
+- metacal was reconvolving by the pixel *twice*, which resulted in larger
+  reconvolved PSFs and thus somewhat lower response.  This did not cause
+  any bias.
 - fixed computation of Jacobian center for `ngmix.KObservation`s with mixed
   even-odd dimensions
 - the total shape amplitude `g` is now reset properly when using the
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -36,8 +36,8 @@
 ### bug fixes
 
 - metacal was reconvolving by the pixel *twice*, which resulted in larger
-  reconvolved PSFs and thus somewhat lower response.  This did not cause
-  any bias.
+  reconvolved PSFs and thus somewhat lower, but still accurate, response than
+  the pixel was included once.  This did not cause any bias.
 - fixed computation of Jacobian center for `ngmix.KObservation`s with mixed
   even-odd dimensions
 - the total shape amplitude `g` is now reset properly when using the
--- a/ngmix/metacal/metacal.py
+++ b/ngmix/metacal/metacal.py
@@ -305,6 +305,7 @@ class MetacalDilatePSF(object):
                 ny=ny,
                 wcs=self.image.wcs,
                 dtype=np.float64,
+                method='no_pixel',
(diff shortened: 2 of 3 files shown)


beckermr (changes requested):
I don't think the change log is right. The metacal observations in the returned duct definitely had an image and a psf image that did not quite match. However, metacal shears divide the shape in the noshear image by the response. I think the error in the psf size falls out so that the final shear is correct.

esheldon:
the psf image is only used for calculating statistics, e.g. correlating the galaxy shape with the psf shape etc.  It won't cause any bias

esheldon:
to put it another way: the psf correction happens when we deconvolve by the original psf; an error in the downstream psf is completely decoupled

beckermr:
Exactly. So the statement that the response is incorrect in changelog is not true. The metacal response as defined is totally fine.

esheldon:
I wasn't saying the response was wrong,  I was saying that the response was lower than if we hadn't added the extra pixel convolution.

We could also make the response lower by reconvolving by an arbitrarily bigger PSF, but it would still be correct

beckermr:
Ahhhh I see.  Thanks!

beckermr (approved):

PR #163, opened 2021-06-23
TST run regression tests against 1.3.9

beckermr:
This PR fixes the regression tests.

### DIFF
--- a/.github/workflows/mdet.yml
+++ b/.github/workflows/mdet.yml
@@ -58,9 +58,9 @@ jobs:
           popd
 
           pip uninstall ngmix -y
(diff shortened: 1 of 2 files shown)


beckermr:
@esheldon I had to move the tolerance to 8e-6 from 2e-6 between 1.3.9 and 2.0.0 on master. I don't know if we should be worried about this or not. Seems like a small enough difference that it wouldn't be a big deal, but maybe not?

esheldon:
I agree it seems fine, could be the results have that much noise in them.

esheldon (approved):

esheldon:
I'm not claiming I know where that noise is coming from

PR #167, opened 2021-08-31
add fixed covariance em

esheldon:


esheldon:
An em algorithm where the covariance of the gaussians is held fixed

PR #195, opened 2021-11-01
BUG fix buggy admom usage in metacal

esheldon:
In MetacalFitGaussPSF when using adaptive moments to get the psf
     model, be careful to use the fitted parameters rather than going
     through an intermediate gmix, which converts e1, e2 to g1, g2 and
     can fail in rare cases.


beckermr (approved):

beckermr:
thank you @esheldon!

PR #199, opened 2021-11-01
No longer require rng forMetacalGaussPSF and MetacalAnalyticPSF

### LINKED ISSUE #172: Remove rng requirement for `MetacalGaussPSF, MetacalAnalyticPSF`
this was to help downstream code using ngmix gaussian fitter, which can fail for a noiseless gaussian image as input.

But downstream can use admom or add noise themselves

On the other hand some codes may break

esheldon:
closes #172 

   It is no longer required to send a random number generator
   to the constructors for MetacalGaussPSF and MetacalAnalyticPSF.  If one
   is sent it is still used to add a small amount of noise.

### DIFF
--- a/ngmix/metacal/convenience.py
+++ b/ngmix/metacal/convenience.py
@@ -51,9 +51,9 @@ def get_all_metacal(
         the sheared, correlated noise component.  Default True
     rng: np.random.RandomState
         A random number generator; this is required if fixnoise is True and
-        use_noise_image is False.  It is also required when psf= is sent, in
-        order to add a small amount of noise to the rendered image of the
-        psf.
+        use_noise_image is False.  If the keyword psf= is sent, it will be used
+        to to add a small amount of noise to the rendered image of the psf,
+        but is not required.
     use_noise_image: bool, optional
         If set to True, use the .noise attribute of the observation
         for fixing the noise when fixnoise=True.
--- a/ngmix/metacal/convenience.py
+++ b/ngmix/metacal/convenience.py
@@ -53,7 +53,7 @@ def get_all_metacal(
         A random number generator; this is required if fixnoise is True and
         use_noise_image is False, or if psf is set to 'fitgauss'.
 
-        For psf is set to 'gauss', 'fitgauss' or a galsim object, it will be
+        If the psf is set to 'gauss', 'fitgauss' or a galsim object, it will be
         used to to add a small amount of noise to the rendered image of the psf
     use_noise_image: bool, optional
         If set to True, use the .noise attribute of the observation
--- a/ngmix/metacal/convenience.py
+++ b/ngmix/metacal/convenience.py
@@ -53,9 +53,8 @@ def get_all_metacal(
         A random number generator; this is required if fixnoise is True and
         use_noise_image is False, or if psf is set to 'fitgauss'.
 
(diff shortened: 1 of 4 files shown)


ngmix/metacal/convenience.py
```diff
@@ -51,9 +51,9 @@ def get_all_metacal(
         the sheared, correlated noise component.  Default True
     rng: np.random.RandomState
         A random number generator; this is required if fixnoise is True and
-        use_noise_image is False.  It is also required when psf= is sent, in
-        order to add a small amount of noise to the rendered image of the
-        psf.
+        use_noise_image is False.  If the keyword psf= is sent, it will be used
```
beckermr, line 54: This is confusing since psf is always sent.

beckermr: I think it applies noise if the RNG is not None.

ngmix/tests/test_metacal.py
```diff
@@ -42,6 +42,40 @@ def test_metacal_smoke(psf):
...
+
+    kw = {
+        'psf': psf, 'fixnoise': False, 'types': ['noshear', '1p', '1m'],
+    }
+    if send_rng:
+        kw['rng'] = rng
+
+    if psf == 'fitgauss' and not send_rng:
+        with pytest.raises(ValueError):
+            obs_dict1 = ngmix.metacal.get_all_metacal(obs, **kw)
```
beckermr, line 63: I thought we never needed the RNG. Is this not true?

beckermr (commented):
I am confused by the change TBH.

ngmix/tests/test_metacal.py
esheldon: It is needed for fitgauss to generate guesses for the fitter

ngmix/metacal/convenience.py
esheldon: fixed I think

ngmix/metacal/convenience.py
```diff
@@ -51,9 +51,10 @@ def get_all_metacal(
         the sheared, correlated noise component.  Default True
     rng: np.random.RandomState
         A random number generator; this is required if fixnoise is True and
-        use_noise_image is False.  It is also required when psf= is sent, in
-        order to add a small amount of noise to the rendered image of the
-        psf.
+        use_noise_image is False, or if psf is set to 'fitgauss'.
+
+        For psf is set to 'gauss', 'fitgauss' or a galsim object, it will be
```
beckermr, line 56:
```suggestion
        For psf set to 'gauss', 'fitgauss' or a galsim object, it will be
```

ngmix/metacal/metacal.py
```diff
@@ -590,8 +593,8 @@ class MetacalFitGaussPSF(MetacalGaussPSF):
     obs: ngmix.Observation
         Observation on which to run metacal
     rng: numpy.random.RandomState
-        Random number generator for adding a small amount of noise to the
-        gaussian psf image
+        Optional random number generator for adding a small amount of noise to
+        the gaussian psf image
```
beckermr, line 597: This should say required, right?

beckermr (commented):
docs need a bit of TLC

beckermr (approved):

PR #206, opened 2021-11-15
Implement copy(), copy/deepcopy and __eq__ for jacobians, observations, gmix, as well as get_data for jacobians

esheldon:


esheldon:
while I'm at it I'm going to add some more of these for other ngmix data types

beckermr (changes requested):

esheldon:
I'm stumped currently on this infinite recursion

beckermr:
Hmmmm.

The docs say:

```
In order for a class to define its own copy implementation, it can define special methods __copy__() and __deepcopy__(). The former is called to implement the shallow copy operation; no additional arguments are passed. The latter is called to implement the deep copy operation; it is passed one argument, the memo dictionary. If the __deepcopy__() implementation needs to make a deep copy of a component, it should call the deepcopy() function with the component as first argument and the memo dictionary as second argument. The memo dictionary should be treated as an opaque object.
```

Maybe we need to the memo dict for our object before calling copy and just return any copy we have already made?

We also might want to use the memo dict properly per the docs.

esheldon:
the id won't have changed, so we need to replace it in memo

esheldon:
I mean if, say the arrays change

ngmix/observation.py
```diff
@@ -979,6 +1054,34 @@ def get_s2n_sums(self):
...
+        for obslist in self:
+            new_mbobs.append(obslist.copy(memo=memo))
+
+        return new_mbobs
+
+    def __copy__(self):
+        return self.copy()
+
+    def __deepcopy__(self, memo):
+        result = self.copy()
```
beckermr, line 1072:
```suggestion
        result = self.copy(memo=memo)
```

beckermr:
Huzzah!

beckermr (approved):

beckermr:
nice work!

esheldon:
thanks for the excellent review

PR #208, opened 2021-11-22
filter NumbaExperimentalFeatureWarning

esheldon:
We've been using this for years, but recently numba started
issuing a warning about it!


esheldon:
this is feature has been available since ngmix was converted from C to numba, about 8 years ago.  WE should probably bring this up with the numba devs

esheldon:
this spurious warning is gone with the latest numba

PR #210, opened 2021-11-25
ENH apodize the pre-PSF moments

beckermr:
This PR adds apodization for the pre-PSF moments and tests that it does the right thing.

### DIFF
--- a/ngmix/tests/test_prepsfmom.py
+++ b/ngmix/tests/test_prepsfmom.py
@@ -2,7 +2,7 @@ import galsim
 import numpy as np
 import pytest
 
-from ngmix.prepsfmom import KSigmaMom, PGaussMom
+from ngmix.prepsfmom import KSigmaMom, PGaussMom, _build_square_apodization_mask
 from ngmix import Jacobian
 from ngmix import Observation
 from ngmix.moments import make_mom_result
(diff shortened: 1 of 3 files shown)


ngmix/tests/test_prepsfmom.py
```diff
@@ -721,3 +721,109 @@ def test_prepsfmom_comp_to_gaussmom(
...
+    )
+
+    from ngmix.gaussmom import GaussMom
+    res_gmom = GaussMom(fwhm=mom_fwhm).go(obs=obs_ap)
+
+    return res, res_gmom
+
+
+@pytest.mark.parametrize("flux_factor", [1e2, 1e3, 1e5])
+def test_prepsfmom_apodize(flux_factor):
```
esheldon, line 814: They are better, but how good are they?  Does the zero case have bad ringing, and is the resulting moments very bad, but not so for the apodized?

I wonder if the ringing would also produce spurious detections.

esheldon (changes requested):

ngmix/tests/test_prepsfmom.py
beckermr: The differences can be significant for stamps with bright sources on the edge. 

I don't get the spurious detection question. We aren't detecting anything here.

esheldon: Yeah, I know you aren't detecting but I'm curious if one would see spurious detections if one was detecting.  Anyway, out of scope for this.

esheldon (approved):

PR #212, opened 2021-12-09
BUG add kind attribute for PrePSFMom fitter

beckermr:
This PR sets the kind attribute for the prePSFMom fitter.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,4 @@
-## v2.0.6 (unreleased)
+## v2.0.6
 
 ### bug fixes
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,10 @@
+## v2.0.6 (unreleased)
+
+### bug fixes
+
+    - Fixed a bug where the `kind` attribute was not set for the `PrePSFMom` fitter.
+
+
 ## v2.0.5
 
 ### new features
--- a/ngmix/prepsfmom.py
+++ b/ngmix/prepsfmom.py
@@ -193,9 +193,6 @@ class KSigmaMom(PrePSFMom):
         The apodization radius for the stamp in pixels. The default of 1.5 is likely
         fine for most ground based surveys.
     """
-
-    kind = "ksigma"
-
     def __init__(self, fwhm, pad_factor=4, ap_rad=1.5):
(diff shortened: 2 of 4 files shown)


CHANGES.md
```diff
@@ -1,3 +1,10 @@
+## v2.0.6 (unreleased)
```
beckermr, line 1:
```suggestion
## v2.0.6
```

ngmix/prepsfmom.py
```diff
@@ -42,6 +42,14 @@ def __init__(self, fwhm, kernel, pad_factor=4, ap_rad=1.5):
         self.pad_factor = pad_factor
         self.kernel = kernel
         self.ap_rad = ap_rad
+        if self.kernel == "ksigma":
+            self.kind = "ksigma"
```
esheldon, line 46: I think this makes the `kind` set in the inheriting classes redundant

esheldon (changes requested):

ngmix/prepsfmom.py
esheldon: Also not this makes it impossible to do inheritance, since the type won't be "registered' 

I'm not sure what the right thing to do is.

erykoff: Oh, you're saying that in the inherited classes it's a class variable, but this setting is an instance variable?

beckermr: > Also not this makes it impossible to do inheritance, since the type won't be "registered'

What python foo is this?

esheldon: Really all I meant is what Eli said.

esheldon: My point about inheritance was just that these two options have special status ksigma and pgauss, and if you tried to make your own the base class would not know about it.

beckermr: Ahhh right. In our case the baee class errors and you need to adjust the guts. The inheritance here is more about the interface then extensibility.

beckermr: Probably this but I'm not gonna go down this path. 

https://stackoverflow.com/questions/45400284/understanding-init-subclass

esheldon: meh

esheldon (approved):

PR #216, opened 2022-03-18
PERF cache FFTs of various quantities

beckermr:
This PR adds caching of FFTs of various images in order to speed up the code.

### DIFF
--- a/ngmix/tests/test_prepsfmom.py
+++ b/ngmix/tests/test_prepsfmom.py
@@ -7,6 +7,8 @@ from ngmix.prepsfmom import (
     KSigmaMom, PGaussMom,
     _build_square_apodization_mask,
     PrePSFMom,
+    _gauss_kernels,
+    _zero_pad_and_compute_fft_cached_impl,
 )
 from ngmix import Jacobian
 from ngmix import Observation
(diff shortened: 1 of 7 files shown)


beckermr:
This one is ready for review @esheldon!

esheldon (changes requested):
We should also have a speedup test for metacal, asserting it is faster by X, to justify the change.

ngmix/tests/test_prepsfmom.py
```diff
@@ -94,6 +98,135 @@ def test_prepsfmom_raises_badjacob(cls):
...
+
+    print("%0.4f ms per fit" % (dt3/nfit*1000))
+
+    # we should never miss again for the calls above
+    assert _gauss_kernels.cache_info().misses == 2
+    assert _zero_pad_and_compute_fft_cached_impl.cache_info().misses == 4 + nfit
+
+    # if numba stuff is cached this does not work so commented out
+    # assert dt2 < dt1
+    assert dt3/nfit < dt2
```
beckermr, line 227:
```suggestion
    assert dt3/nfit < dt2*0.6
```

esheldon:
what about for metacal?  Is there an important and predictable speedup for this caching?

beckermr:
The speedup comes from caching the galsim interpolated image object of the original image. It's FFT gets cached internally by galsim and this saves a bunch of time in generating the different metacal images at each color.

beckermr:
I need to increase the digits on the jacobian repr for this one.

beckermr:
ok @esheldon This PR is ready to go. I added a test showing the ~30% gains from caching for mdet-sized images.

esheldon (approved):

PR #220, opened 2022-07-21
BUG use area norm for all gauss mom things

beckermr:
This PR changes GaussMom to use the 1/area normalization consistently everywhere in its outputs.


beckermr:
@esheldon We should work on this one first.

beckermr:
Note that we could change this directly in the outputs of the weighted moments from GMix but that would be more breaking.

esheldon:
I think this is harmless since we always take ratios

PR #227, opened 2022-08-24
PERF add O(N) phase shift computation

beckermr:
Trying some stuff here after talking with Mike.

### DIFF
--- a/ngmix/prepsfmom.py
+++ b/ngmix/prepsfmom.py
@@ -276,6 +276,23 @@ def _measure_moments_fft(
         cen_phase = _compute_cen_phase_shift(drow, dcol, dim, msk=msk)
         kim *= cen_phase
 
+    # we only sum where the kernel is nonzero
+    fkf = kernels["fkf"]
+    fkr = kernels["fkr"]
+    fkp = kernels["fkp"]
+    fkc = kernels["fkc"]
+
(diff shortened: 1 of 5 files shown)


ngmix/prepsfmom.py
```diff
@@ -276,6 +276,23 @@ def _measure_moments_fft(
         cen_phase = _compute_cen_phase_shift(drow, dcol, dim, msk=msk)
         kim *= cen_phase
 
+    # we only sum where the kernel is nonzero
```
beckermr, line 279:
```suggestion
```

ngmix/prepsfmom.py
```diff
@@ -285,13 +301,6 @@ def _measure_moments_fft(
...
     df4 = df2 * df2
 
-    # we only sum where the kernel is nonzero
-    fkf = kernels["fkf"]
-    fkr = kernels["fkr"]
-    fkp = kernels["fkp"]
-    fkc = kernels["fkc"]
-
-    mom_norm = kernels["fk00"]
     mf = np.sum((kim * fkf).real) * df2
```
beckermr, line 304:
```suggestion
    # we only sum where the kernel is nonzero
    mf = np.sum((kim * fkf).real) * df2
```

PR #232, opened 2022-11-14
Pass on ignore_zero_weight to medsreaders.  Fill in AM missing pixels

esheldon:


esheldon:
This PR has two parts.

The first is to expose the ignore_zero_weight to the medsreaders, so we can create Observations with zero weights or uberseg with these zero weight pixels in the pixels array.  This faciliates filling in the pixels in AM

Second is to 1) calculate a flux for the current AM model and 2) fill in zero weight pixels using the properly normalized model.

I also am now using this flux/flux error and flags in the admom result

beckermr (changes requested):
I think the change to Admom filling in pixels should be optional and set when you make the fitter. Otherwise we are pushing a breaking change that cannot be reversed later.

CHANGES.md
```diff
@@ -1,3 +1,13 @@
+## v2.3.1
+
+### New Features
+
+    - Adaptive moments now calculate the flux.  The flux is a template
+      flux rather than the weighted flux, and thus is more comparable to the flux
+      from a maximum likelyhood fit rather than from weighted moments.
```
beckermr, line 7: I don't understand this comment at all, and I think maybe it is wrong.

ngmix/tests/test_admom.py
```diff
@@ -208,45 +207,130 @@ def test_admom_find_cen(g1_true, g2_true, snr):
     assert np.abs(dcol_mean)/dcol_err < 5, (dcol_mean/dcol_err)
 
 
-@pytest.mark.parametrize('pixel_scale', [0.25, 0.125])
-@pytest.mark.parametrize('fwhm', [2, 0.5])
-@pytest.mark.parametrize('image_size', [53])
-@pytest.mark.parametrize('mom_fwhm', [2, 2.5])
-def test_admom_comp_to_gaussmom_flux(image_size, fwhm, pixel_scale, mom_fwhm):
```
beckermr, line 215: Removing this test makes it hard to figure out if the flux normalization is different between the two versions.

beckermr: At minimum, a test comparing to a Gaussian fit would be clarifying.

esheldon:
Filing only occurs when there are zero weight pixels, which is not the default for Observations.  No examples in the /examples director showed using it.  

Anyone using admom with masked data in the past was getting a biased result.

ngmix/tests/test_admom.py
esheldon: The new test compares to the true input flux.

beckermr:
Right but I want to try symmetrized zero-weight pixels. I can't with this PR.

esheldon:
Yes you can, just construct the Observation with the default `ignore_zero_weight=True`

beckermr:
Ah I see. Thank you.

CHANGES.md
```diff
@@ -1,3 +1,13 @@
+## v2.3.1
+
+### New Features
+
+    - Adaptive moments now calculate the flux of the model rather than a weighted
+      aperture flux.  The flux is a template flux (same method as
```
beckermr, line 6: I still don't understand what you mean by this:

```
Adaptive moments now calculate the flux of the model rather than a weighted aperture flux. 
```

Sorry, I am not trying be dense. The flux function below is using the data in the pixels array which is not the model.

```
       if pixel['ierr'] <= 0:
            val = model * oflux
        else:
            val = pixel['val']

        xcorr_sum += model * val * ivar
        msq_sum += model * model * ivar

    flux = xcorr_sum / msq_sum

    res['flux'] = flux
```

I don't see how this quantity differs from index 5 of momsums that admom normally computes.

beckermr: Sorry I should say differs from index 5 of momsums after the correction factors that were put into the code before.

beckermr: A while back admom did return the model normalization iirc but I put in a PR to change that to make it match the other moments fitters. I think what is happening here is that you have refactored the code a bit and correctly identified that I should have made a bigger deal about this than I did.

esheldon: This gives the best fit flux rather than the gaussian aperture weighted flux, it is equivalent to doing a psf flux style template fit.

See results.PSFFluxFitModel

This is the flux needed to properly normalize the model and fill in missing data, and is equivalent to the flux we would get doing the maximum likelihood fit

beckermr: Yeah but I am looking at the old and new code and they appear to compute the same thing.

beckermr: Ohhhh I missed a factor the weight in the denominator. Sorry.

beckermr (approved):

esheldon:
This results in an extra evaluation of the  model over the pixels during each iteration, so is sub-optimal.  I think we can probably eliminate some of the passes by pre-evaluating the model.

But I first want to test how this works in some simulations before optimizing.

ngmix/admom/admom.py
```diff
@@ -409,11 +409,19 @@ def get_result(ares):
...
+    #     else:
+    #         res['flux_flags'] |= ngmix.flags.NONPOS_VAR
+
+    fc = res['sums_cov'][5, 5]
+    if fc > 0:
+        res['s2n'] = res['sums'][5] / np.sqrt(fc)
+        res['flux_err'] = res['flux'] / res['s2n']
+    else:
+        res['flux_flags'] |= ngmix.flags.NONPOS_VAR
+        res['flags'] |= ngmix.flags.NONPOS_VAR
```
beckermr, line 424: Be careful here. The old sums[5] entry needs some factors to make its normalization match other things properly. 

See this comment I had

```
# this is a fun set of factors
            # jacobian area is because ngmix works in flux units
            # the wgt_norm and wsum compute the weighted flux and normalize
            # the weight kernel to peak at 1
            fnorm = jac_area * wgt_norm * res["wsum"]
            res['flux'] = res['sums'][5] / fnorm
```

I think what you have works but it might be worth leaving some comments for future folks.

esheldon: thanks, this is temporary

PR #235, opened 2022-12-10
add seg to Observation, read it in medsreaders

### LINKED ISSUE #233: Add optional seg for Observations and read with medsreaders
sometimes we want to know if a pixel was assigned to an object

esheldon:
closes #233

### DIFF
--- a/ngmix/medsreaders.py
+++ b/ngmix/medsreaders.py
@@ -224,6 +224,11 @@ class NGMixMEDS(_MEDS):
         except Exception:
             mfrac = None
 
+        try:
+            seg = self.get_cutout(iobj, icutout, type='seg')
+        except Exception:
+            seg = None
+
         if weight_type == 'uberseg':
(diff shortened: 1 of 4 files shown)


esheldon:
Is this an issue with the version of metadetect?

beckermr:
Yeah just add det bands to that list.

esheldon:
cseg isn't that useful, I forgot it sets the default value to the id of the object rather than zero

beckermr:
So uberseg in the meds readers always uses the coadd seg map. It looks like this class will default to use the seg on the obs which in principle will be different. Do we care about this?

esheldon:
Are you proposing that the default change if uberseg is used for weight type?

beckermr:
That's one way to make things less confusing possibly. I think all of the weight types that rely on a seg map use the coadd one so we might want to do this generally.

PR #236, opened 2023-10-03
fix bug with metacal psf method "gauss"

esheldon:
The psf determination was being done on the pre pixel rather than post pixel psf.  This required a Convolve with the pixel afterward, which does not work as expected when the wcs is complex

Added a metacal accuracy test as well

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,10 @@
+## v2.3.1
+
+## Bug Fixes
+
+    - The metacal psf reconvolution method 'gauss' (which uses MetacalGaussPSF)
+      now works with a complex WCS
+
 ## v2.3.0
 
 ### new features
--- a/mdet_tests/test_mdet_regression.py
+++ b/mdet_tests/test_mdet_regression.py
@@ -268,7 +268,7 @@ def test_mdet_regression(fname, write=False):
                     ),
                 }
             else:
-                assert col in ["shear", "shear_bands"]
+                assert col in ["det_bands", "shear", "shear_bands"]
 
 
 if __name__ == "__main__":
--- a/ngmix/_version.py
+++ b/ngmix/_version.py
@@ -1,1 +1,1 @@
-__version__ = '2.3.0'  # noqa
+__version__ = '2.3.1'  # noqa
--- a/ngmix/metacal/metacal.py
+++ b/ngmix/metacal/metacal.py
@@ -419,6 +419,7 @@ class MetacalDilatePSF(object):
         # interpolated psf deconvolved from pixel.  This is what
         # we dilate, shear, etc and reconvolve the image by
         self.psf_int_nopix = galsim.Convolve([psf_int, self.pixel_inv])
+        self.psf_int = psf_int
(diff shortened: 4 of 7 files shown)


esheldon:
@beckermr is that failure obvious to you?

beckermr:
I'll check the test thing in a bit. I still don't get why this fails or what is going on.

esheldon:
Looks like the column `det_bands` was not expected

beckermr (changes requested):
So I am looking at this in detail and this code is wrong IMHO. IDK why it works, but it is not logically consistent. 

The function _get_gauss_target_psf returns a `galsim.Gaussian` object. Then `do_dilate` applies a dilation based on the shear. Finally, the output of `_get_dilated_psf` is used in this block of code

```python
        if key not in self._psf_cache:
            psf_grown = self._get_dilated_psf(shear, doshear=doshear)

            # this should carry over the wcs
            psf_grown_image = self.psf_image.copy()

            try:
                psf_grown.drawImage(
                    image=psf_grown_image,
                    method='no_pixel',  # pixel is already in psf
                )
            except RuntimeError as err:
                # argh, galsim uses generic exceptions
                raise GMixRangeError("galsim error: '%s'" % str(err))

            self._psf_cache[key] = (psf_grown_image, psf_grown)
```

Notice here the image is rendered with `no_pixel`. However, with the changes above, the pixel was never added back in. It doesn't matter that you kept the pixel when calling `_get_gauss_target_psf` since this function simply returns a raw galsim object that knows nothing of the pixel.

What I think is happening is that indeed the pixel handling has gone wrong and the change here happens to make things just right for the test case.

esheldon:
With the new code the pixel is never taken out

esheldon:
There was never a reason to take out the pixel before running that algorithm.  In fact I think it was a bad idea to do so because the determination of the gaussian was affected by noise amplification.

beckermr:
Sure but you do have to add the pixel back in in order for the rendering to be correct with `no_pixel`.

beckermr:
Not taking it out doesn't matter. The returned reconv PSF is not an interpolated image. It is a Gaussian from galsim.

esheldon:
This is the same thing we do with fitgauss.  In that case we fit the pixelized PSF, dilate it and render with no pixel.

It is true that there could be problems with an undersampled PSF, but that's not unique to this method.

esheldon:
Right, when I said the pixel is in it I meant that it is broader than the pixelized PSF we started with, so should be valid.

But its not correct that the algorithm is just returning a too small PSF.  Any time I reconvolve by the pixel I'm getting a bias.

For example I can just pick some good PSF by hand, say the one returned by fitgauss.  If I convolve that by the pixel and use it I get a bias.  This is what is going wrong.

beckermr:
Ahhhhhh. With a WCS that is not diagonal, the pixel is asymmetric. It also not true that the pixel is symmetric even with a diagonal WCS. I am betting we have an additive effect that looks like a multiplicative one.

PR #241, opened 2024-06-14
add rho4 to adaptive moments

### LINKED ISSUE #240: Implement 4th order moments in admom

esheldon:
closes #240 

Add some docs as well and remove unused
entry in admom dtype

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -6,7 +6,6 @@ omit =
     ngmix/admom/*_nb.py
     ngmix/gmix/*_nb.py
     ngmix/pixels/*_nb.py
-    ngmix/tests/*
 
 [report]
 exclude_lines = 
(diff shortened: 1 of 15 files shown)


.coveragerc
```diff
@@ -6,6 +6,7 @@ omit =
     ngmix/admom/*_nb.py
     ngmix/gmix/*_nb.py
     ngmix/pixels/*_nb.py
+    ngmix/tests/*
```
beckermr, line 9: It is a good idea to include tests in coverage since they should be at 100% to ensure we actually execute all of them.

esheldon:
Looks like the speed test is failing, @mrb do you think we could tweak that test?

beckermr:
Yeah raise the 0.6 to 0.8

PR #248, opened 2025-03-21
Compatibility with numpy 2

arunkannawadi:
This PR makes `ngmix` compatible with `numpy>=2` and `numpy<2`.

The recommended method is to replace `np.array(..., copy=False)` with `np.asarray` [link](https://numpy.org/devdocs/numpy_2_0_migration_guide.html#adapting-to-changes-in-the-copy-keyword). However, given the necessity of `ndmin` which is not available in `asarray`, I took the [scipy approach](https://github.com/scipy/scipy/pull/20172) here.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,7 +11,7 @@ jobs:
     name: tests
     strategy:
       matrix:
-        pyver: ["3.9", "3.10", "3.11"]
+        pyver: ["3.9", "3.10", "3.11", "3.12"]
 
     runs-on: "ubuntu-latest"
 
--- a/ngmix/defaults.py
+++ b/ngmix/defaults.py
@@ -1,5 +1,10 @@
 import numpy as np
 
+# Shim to support numpy >= 2 and < 2.0.0.
+if np.lib.NumpyVersion(np.__version__) >= "2.0.0":
+    copy_if_needed = None
+else:
+    copy_if_needed = False
 
 # default values
 PDEF = -9.999e9  # parameter defaults
--- a/ngmix/flags.py
+++ b/ngmix/flags.py
@@ -81,7 +81,7 @@ def get_flags_str(val, name_map=None):
     if val < 0:
         raise ValueError(f"Flag value {val} must be non-negative.")
 
-    # Cast to uint64 to ensure to allow for maximal flexibility.
+    # Cast to uint32 is sufficient given the range of flags.
     # This is because the second argument to the bitwise AND operator (&) below
     # would get implicitly cast to the same type as `val`.
     val = np.array(val, dtype=np.uint32)
--- a/ngmix/flags.py
+++ b/ngmix/flags.py
@@ -63,7 +63,7 @@ def get_flags_str(val, name_map=None):
     Parameters
     ----------
     val : int
-        The flag value.
+        The flag value. This must be in the range [0, 2**32).
     name_map : dict, optional
         A dictionary mapping values to names. Default is global at
         ngmix.flags.NAME_MAP.
--- a/ngmix/flags.py
+++ b/ngmix/flags.py
@@ -63,7 +63,7 @@ def get_flags_str(val, name_map=None):
     Parameters
     ----------
     val : int
-        The flag value. This must be in the range [0, 2**32).
+        The flag value. This must be non-negative.
     name_map : dict, optional
         A dictionary mapping values to names. Default is global at
         ngmix.flags.NAME_MAP.
@@ -76,14 +76,20 @@ def get_flags_str(val, name_map=None):
     if name_map is None:
         name_map = NAME_MAP
 
-    if val < 0 or val >= 2**32:
-        import warnings
-        warnings.warn(f"Flag value {val} is out of range [0, 2**32).")
+    # Negative values are always invalid and cause confusion if cast to an
+    # unsigned integer.
+    if val < 0:
+        raise ValueError(f"Flag value {val} must be non-negative.")
+
+    # Cast to uint64 to ensure to allow for maximal flexibility.
+    # This is because the second argument to the bitwise AND operator (&) below
+    # would get implicitly cast to the same type as `val`.
+    val = np.array(val, dtype=np.uint64)
 
     nstrs = []
     for pow in range(31):
         fval = 2**pow
-        if ((np.array(val).astype(np.uint32) & fval) != 0):
+        if ((val & fval) != 0):
             if fval in name_map:
                 nstrs.append(name_map[fval])
             else:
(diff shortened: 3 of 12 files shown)


.github/workflows/test.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     name: tests
     strategy:
       matrix:
-        pyver: ["3.9", "3.10", "3.11"]
+        pyver: ["3.9", "3.10", "3.11", "3.12"]
```
arunkannawadi, line 14: I added 3.12 to explicitly check this with the python version in the latest `rubin-env`. I left 3.9 as is, since it acts as an implicit check with `numpy<2`.

ngmix/defaults.py
```diff
@@ -1,5 +1,10 @@
 import numpy as np
 
+# Shim to support numpy >= 2 and < 2.0.0.
+if np.lib.NumpyVersion(np.__version__) >= "2.0.0":
+    copy_if_needed = None
+else:
+    copy_if_needed = False
```
arunkannawadi, line 7: `defaults` may not be the best place for this, but happy to move it wherever it might be more obvious for you.

ngmix/flags.py
```diff
@@ -77,7 +79,7 @@ def get_flags_str(val, name_map=None):
     nstrs = []
     for pow in range(32):
         fval = 2**pow
-        if ((val & fval) != 0):
+        if ((np.uint32(val) & fval) != 0):
```
arunkannawadi, line 82: With `numpy>=2`, this now results in an OverflowError. We could either adopt limiting `pow` to `range(30)` as we did in [metadetect here](https://github.com/esheldon/metadetect/pull/224). I figured it's probably safer to cast it instead since that leaves the most-significant bits still usable.

beckermr: We should 

- limit the loop to 31
- raise a warning if the input value is bigger than 2**32-1 or < 0
- safely cast using `np.array(value).astype(dtype)`

A direct call to `np.uint32` is raising a warning that it will raise an error in the future for my local copy of numpy at 1.26.4.

.github/workflows/test.yml
beckermr: SGTM!

ngmix/flags.py
arunkannawadi: Oh, I didn't know there were going to be future releases of `numpy 1.26` that would convert this to an error.

beckermr: They might mean in numpy 2. IDK.

ngmix/flags.py
```diff
@@ -74,10 +76,14 @@ def get_flags_str(val, name_map=None):
...
+    if val < 0 or val >= 2**32:
+        import warnings
+        warnings.warn(f"Flag value {val} is out of range [0, 2**32).")
+
     nstrs = []
-    for pow in range(32):
+    for pow in range(31):
         fval = 2**pow
-        if ((val & fval) != 0):
+        if ((np.array(val).astype(np.uint32) & fval) != 0):
```
beckermr, line 86: Let's move this cast up to the top and do it once.

arunkannawadi: I want to make sure we don't do too many redundant things. If we limit the loop to `range(31)`, then casting isn't even required. The error was arising from implicit casting of `fval` and not `val`. We should perhaps raise a `ValueError` if `val` is outside the specific range to avoid any underflow errors be done with it. For instance, if `val = np.int32(-5)`, we'd get a completely wrong result with just a seemingly benign warning.

arunkannawadi: My suggestion is to modify your list above to the following:

We should

- limit the loop to 31
- raise an exception if the input value is bigger than 2**32-1 or < 0

arunkannawadi: OK, in that case, having an upper limit doesn't make sense then. We raise an exception if `val` is negative and cast it to `uint64` safely always.

arunkannawadi: Well, what do you know. The code will now work even if `val` is `np.int8` etc.

arunkannawadi: My plan of always casting to `uint64` doesn't work for `numpy<2`. However, given the range of values that `fval` takes, which is hard coded in this routine,`uint32` should be sufficient. It does not pose any restriction on the exact integer type of `val` .

arunkannawadi: I'd hate to break anything because I didn't consider a corner case, so I added a small unit test as well.

ngmix/tests/test_flags.py
```diff
@@ -0,0 +1,14 @@
...
+import pytest
+from ..flags import get_flags_str, NAME_MAP, LOW_DET
+
+
+@pytest.mark.parametrize(
+    "dtype", [np.int8, np.uint8, np.int16, np.uint16, np.int32, np.uint32, np.int64, np.uint64]
+)
+def test_get_flags_str(dtype):
+    """Test that get_flags_str works for different integer types."""
+    val = np.array(LOW_DET, dtype=dtype)
```
beckermr, line 13: For the bigger types we should or with 2**45 to ensure things are OK.

arunkannawadi: In `numpy<2`, bitwise or to `uint64` is prohibited for some reason. I'll condition on that if that's okay.

beckermr: Ack I forgot and yes pls do!

ngmix/tests/test_flags.py
```diff
@@ -11,8 +11,6 @@
 def test_get_flags_str(dtype):
     """Test that get_flags_str works for different integer types."""
     # Set a high-enough bit in addition to the lower bit to ensure things work.
-    val = np.array(LOW_DET, dtype=dtype)
-    if val.itemsize > 4 and dtype not in (np.uint64, ):
-        val |= 2**45
+    val = np.array(LOW_DET | 2**45).astype(dtype)
```
arunkannawadi, line 14: Alright, this looks much cleaned than the test with all the conditions.

beckermr: Yeah it just leaves the question of how numpy handles truncating things. That's ok I guess but maybe a comment would be good.

beckermr:
@esheldon comments here or shall we merge?

esheldon (approved):

esheldon:
Arun are you still making additions?

arunkannawadi:
Nope, I don't have any more to add unless there are requests. Could you also please cut a release after merging this?

arunkannawadi:
I shall, but I also notice that changes from #246 have not made it into a release yet. I will let you comment on that. This would bump up the minor version, right?

esheldon:
I'll add notes for that.  I'll do the version bump when I make my changes.

esheldon:
v2.4.0 is released

arunkannawadi:
Thank you, Erin!

Issue #250, opened 2025-10-10
Metadetect won't run on only `noshear`


sidneymau:
I was trying to do some quick metadetect tests in which _only_ the `noshear` step is processed. However, when specifying this in the config, metadetect keeps processing both `noshear` and `1p`. After tracing through the different calls, I suspect this is the cause: https://github.com/esheldon/ngmix/blob/d08b471f4c4d5887df9f9f2551efaf8f2e226150/ngmix/metacal/metacal.py#L146-L150

It would be very useful if I could run through the metadetect code (including detection & fitting) for just `noshear` for same tests of the detection & measurement pipeline, so I wonder if this can be updated?

sidneymau:
I think the only change that would be necessary is to extract out the "get_unsheared" bit from https://github.com/esheldon/ngmix/blob/d08b471f4c4d5887df9f9f2551efaf8f2e226150/ngmix/metacal/metacal.py#L181-L185 and rather have the code run each requested type in sequence (rather than trying to do two at once in this unusual case -- or only doing this is if another shear type has been requested)

PR #251, opened 2025-10-10
Allow independent handling of noshear in metacal

sidneymau:
Proposed solution for #250. Rather than forcing the processing of `1p` if `noshear` is specified, this updates the code to check
1. if _only_ `noshear` is requested, then process that independently
2. if `noshear` and other types are requested, if `noshear` has not yet been run, process that alongside another type

The code produces the desired behavior but there are no tests yet. I'm happy to implement one (probably something that quickly runs metacal with a few different types requests and checks that the output dict has the same types as requested)

### DIFF
--- a/ngmix/metacal/metacal.py
+++ b/ngmix/metacal/metacal.py
@@ -165,7 +165,9 @@ class MetacalDilatePSF(object):
         for type in types:
             sh = shdict[type]
 
-            if (type == 'noshear') and (len(types) == 1):
+            if (type == 'noshear') and (
+                len([type for type in types if 'psf' not in type]) == 1
+            ):
                 # only need noshear
                 _newpsf_image, _newpsf_obj = self.get_target_psf(sh, 'gal_shear')
                 _unsheared_image = self.get_target_image(_newpsf_obj, shear=None)
--- a/ngmix/metacal/metacal.py
+++ b/ngmix/metacal/metacal.py
@@ -143,15 +143,11 @@ class MetacalDilatePSF(object):
             for t in types:
                 assert t in METACAL_TYPES, 'bad metacal type: %s' % t
 
-        # we add 1p here if we want noshear since we get both of those
-        # at once below
-
-        if 'noshear' in types and '1p' not in types:
-            types.append('1p')
-
         shdict = {}
 
         # galshear keys
+        shdict['noshear'] = Shape(0.0, 0.0)
+
         shdict['1m'] = Shape(-step, 0.0)
         shdict['1p'] = Shape(+step, 0.0)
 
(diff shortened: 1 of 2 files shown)


sidneymau:
There may need to be some additional handling in case the requested types include `noshear` and psf types but not normal shear types (in which case `len(types) > 1` would be `True`, but there would never be an opportunity to process `noshear`)

ngmix/metacal/metacal.py
```diff
@@ -167,16 +163,21 @@ def get_all(self, step=DEFAULT_STEP, types=None):
...
             sh = shdict[type]
 
+            if (type == 'noshear') and (
+                len([type for type in types if 'psf' not in type]) == 1
+            ):
+                # only need noshear
+                _newpsf_image, _newpsf_obj = self.get_target_psf(sh, 'gal_shear')
+                _unsheared_image = self.get_target_image(_newpsf_obj, shear=None)
+                obs = self._make_obs(_unsheared_image, _newpsf_image)
+                obs.psf.galsim_obj = _newpsf_obj
```
sidneymau, line 175: The code here reproduces some of the internals of get_obs_galshear but extracting the noshear branch. Maybe the function can be broken up so we things stay consistent with any future changes?

sidneymau:
My bad, didn't catch the existing test. Updated accordingly

esheldon:
Hi Sid. I appreciate the PR.  However, this is a breaking change.

What's your motivation for this change?

sidneymau:
There are a few motivations:
- It would be convenient for testing the detection & fitting pipeline from metadetect in a simplified case. For example, I want to compare the detections and fit fluxes to a truth catalog for some simulated image. This can be accomplished with just the `noshear` branch, and running other shears adds substantial compute time (the limiting step is the metacal process, so it takes longer for each shear type processed)
- It seemed weird to have different shear types processed than were requested, and
- If shear other than `1p` were requested alongside `noshear`, it would be more efficient to process `noshear` alongside that other shear rather than also adding `1p` (e.g., if the user wanted to process `[noshear, 2p, 2m]` then it would be inefficient to process 4 types instead of 3 types)

So the motivations are mostly efficiency with some aesthetic

edit: I also first thought I had a bug in my code because I was getting a different set of shear types in my output than I requested in my config, so I had to spend some time debugging before I realized that this was default (but undocumented, as far as I can tell) behavior

sidneymau:
The alternative I was coming up with was to reconstruct the `MEDSifier`  to get the detection stage run on the `mbobs` but then I may be missing some of the PSF transformations and running the fitters seemed more complicated

esheldon:
You can call `get_obs_galshear` directly.

esheldon:
I guess if you are using `get_all_metacal` it needs `get_all()` method to change

The reason this exists is that noshear/1p can share some calculations, and this improves the speed when you are planning to get more than just noshear.

If anyone relied on that it will break.

sidneymau:
In this case I am running code from `metadetect` (i.e., `metadetect.do_metadetect` and/or using the `Metadetect` class) -- that code runs `get_all_metacal` which runs `get_all`, which is how I arrived here.

Part of my thinking was that noshear can share those calculations with any other shear type, so you can have the calculations be shared between noshear and the first other shear type requested (not just `1p`)

sidneymau:
I do appreciate that it is bad to break standing behavior -- but if the decision is to keep this as-is I do think it would be good to at least document it throughout (none of the docstrings in any of the functions mention this behavior but they will all fall back to it)

beckermr:
It is breaking. I'd be ok supporting a "noshear-only" version that really is only "noshear".

esheldon:
where do we stand on this?  Will add a new api to get noshear only in metadetect.do_metadetect ?

beckermr:
We should add a new api and/or new shear type for this. We should not break the API.

sidneymau:
I don't have a strong feeling about this and happy for this PR to be closed if nobody else finds a pressing need for this.

esheldon:
Yeah, we don't need to change `ngmix`, we can just have `metadetect` call the methods directly for this special case instead of using `get_all_metacal`

PR #256, opened 2026-04-30
Add scale_T() method to GMix objects

esheldon:


esheldon:
This should probably be scale_T rather than set_T to avoid oddness with negatives

ngmix/gmix/gmix.py
```diff
@@ -200,6 +200,32 @@ def get_T(self):
 
         return T
 
+    def set_T(self, T):
```
beckermr: Isn't the math here wrong if the Gaussians are not all centered in the same spot? I think the distances of the Gaussians from the mean position need to be scaled by the ratio or its sqr as well.

esheldon: correct, working on that now

ngmix/tests/test_gmix.py
```diff
@@ -685,3 +685,50 @@ def test_higher_order_nan():
         if key in ['M00', 'MF']:
             continue
         assert np.isnan(res[key])
+
+
+@pytest.mark.parametrize('model', ['gauss', 'turb', 'coellip5'])
+def test_gmix_set_T(model):
```
beckermr: We should add a test for non-cocentered Gaussians.

beckermr (changes requested):
Some comments and questions.

ngmix/gmix/gmix.py
```diff
@@ -200,6 +200,39 @@ def get_T(self):
...
+        Scale the overall T for the Gaussian Mixture
+
+        Parameters
+        ----------
+        scale: float
+            Value to scale the T for the mixture.
+        """
+
+        if scale < 0.0:
+            raise ValueError(f'Requested scale {scale} < 0')
```
beckermr, line 214:
```suggestion
        if scale <= 0.0:
            raise ValueError(f'Requested scale {scale} <= 0')
```

beckermr (changes requested):

ngmix/gmix/gmix.py
esheldon: Should zero be allowed so one can make a delta function?

beckermr: The code divides by the determinant, so I don't think this will work at all.

esheldon: We allow zero determinant so we can fit stars to have zero T.

In the fitter, we don't evaluate the gaussian until after it is convolved by the PSF.

If someone uses it without doing that it will raise an exception, so it is OK.

beckermr (approved):

esheldon:
Thanks for the quick review Matt.

PR #258, opened 2026-06-01
Avoid slowdown in get_flags_str

arunkannawadi:
In #248 , we introduced safe casting to `uint32`, but it seemed to have significantly slowdown the bitwise AND operation that follows. In DP2 production, the calls to `get_flags_str` are taking a significant amount of time, and this is a low-hanging fruit to fix. Along with it is another minor optimization calculating powers of 2 cumulatively. Together, this brings down the per-call time from ~50 μs to ~8 μs, which is a signficant gain given how many times this functions gets called.

### DIFF
--- a/ngmix/flags.py
+++ b/ngmix/flags.py
@@ -84,7 +84,7 @@ def get_flags_str(val, name_map=None):
     # Cast to uint32 is sufficient given the range of flags.
     # This is because the second argument to the bitwise AND operator (&) below
     # would get implicitly cast to the same type as `val`.
-    val = np.array(val, dtype=np.uint32)
+    val = np.asarray([val], dtype=np.uint32)[0]
 
     nstrs = []
     fval = 1
--- a/ngmix/flags.py
+++ b/ngmix/flags.py
@@ -93,7 +93,7 @@ def get_flags_str(val, name_map=None):
         # The bitwise & operation with this mask sets all bits 32 and higher
         # to zero. Thus the resulting int will fit into a unit32 numpy type.
         val = val & 0xFFFFFFFF
-    val = np.asarray(val, dtype=np.uint32)[0]
+    val = np.asarray([val], dtype=np.uint32)[0]
 
     nstrs = []
     fval = 1
--- a/ngmix/flags.py
+++ b/ngmix/flags.py
@@ -63,7 +63,8 @@ def get_flags_str(val, name_map=None):
     Parameters
     ----------
     val : int
-        The flag value. This must be non-negative.
+        The flag value. This must be non-negative. Only the first 32 bits
+        (0 to 31) are considered. Any bits beyond this are ignored.
     name_map : dict, optional
         A dictionary mapping values to names. Default is global at
         ngmix.flags.NAME_MAP.
--- a/ngmix/flags.py
+++ b/ngmix/flags.py
@@ -85,15 +85,7 @@ def get_flags_str(val, name_map=None):
     # Cast to uint32 is sufficient given the range of flags.
     # This is because the second argument to the bitwise AND operator (&) below
     # would get implicitly cast to the same type as `val`.
-    if isinstance(val, int):
-        # Python 3.9 doesn't like to compute bitwise & with uint64 types,
-        # so we only do this for python ints where it is needed.
-        # The value 0xFFFFFFFF is an int with bits 0 to 31 (inclusive)
-        # set to 1 and the rest set to 0.
-        # The bitwise & operation with this mask sets all bits 32 and higher
-        # to zero. Thus the resulting int will fit into a unit32 numpy type.
-        val = val & 0xFFFFFFFF
-    val = np.asarray(val, dtype=np.uint32)
+    val = np.asarray(val).astype(np.uint32)
 
     nstrs = []
     fval = 1
--- a/ngmix/flags.py
+++ b/ngmix/flags.py
@@ -93,7 +93,7 @@ def get_flags_str(val, name_map=None):
         # The bitwise & operation with this mask sets all bits 32 and higher
         # to zero. Thus the resulting int will fit into a unit32 numpy type.
         val = val & 0xFFFFFFFF
-    val = np.asarray([val], dtype=np.uint32)
+    val = np.asarray([val], dtype=np.uint32)[0]
 
     nstrs = []
     fval = 1
--- a/ngmix/flags.py
+++ b/ngmix/flags.py
@@ -85,6 +85,12 @@ def get_flags_str(val, name_map=None):
     # This is because the second argument to the bitwise AND operator (&) below
     # would get implicitly cast to the same type as `val`.
     if isinstance(val, int):
+        # Python 3.9 doesn't like to compute bitwise & with uint64 types,
+        # so we only do this for python ints where it is needed.
+        # The value 0xFFFFFFFF is an int with bits 0 to 31 (inclusive)
+        # set to 1 and the rest set to 0.
+        # The bitwise & operation with this mask sets all bits 32 and higher
+        # to zero. Thus the resulting int will fit into a unit32 numpy type.
         val = val & 0xFFFFFFFF
     val = np.asarray([val], dtype=np.uint32)
 
--- a/ngmix/flags.py
+++ b/ngmix/flags.py
@@ -85,7 +85,15 @@ def get_flags_str(val, name_map=None):
     # Cast to uint32 is sufficient given the range of flags.
     # This is because the second argument to the bitwise AND operator (&) below
     # would get implicitly cast to the same type as `val`.
-    val = np.asarray(val).astype(np.uint32)
+    if isinstance(val, int):
+        # Python 3.9 doesn't like to compute bitwise & with uint64 types,
+        # so we only do this for python ints where it is needed.
+        # The value 0xFFFFFFFF is an int with bits 0 to 31 (inclusive)
+        # set to 1 and the rest set to 0.
+        # The bitwise & operation with this mask sets all bits 32 and higher
+        # to zero. Thus the resulting int will fit into a unit32 numpy type.
+        val = val & 0xFFFFFFFF
+    val = np.asarray(val, dtype=np.uint32)[0]
 
     nstrs = []
     fval = 1
--- a/ngmix/flags.py
+++ b/ngmix/flags.py
@@ -84,9 +84,9 @@ def get_flags_str(val, name_map=None):
     # Cast to uint32 is sufficient given the range of flags.
     # This is because the second argument to the bitwise AND operator (&) below
     # would get implicitly cast to the same type as `val`.
-    if val >= 2**32:
+    if isinstance(val, int):
         val = val & 0xFFFFFFFF
-    val = np.uint32(val)
+    val = np.asarray([val], dtype=np.uint32)
 
     nstrs = []
     fval = 1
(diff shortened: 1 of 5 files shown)


ngmix/flags.py
```diff
@@ -84,14 +84,15 @@ def get_flags_str(val, name_map=None):
     # Cast to uint32 is sufficient given the range of flags.
     # This is because the second argument to the bitwise AND operator (&) below
     # would get implicitly cast to the same type as `val`.
-    val = np.array(val, dtype=np.uint32)
+    val = np.asarray([val], dtype=np.uint32)[0]
```
beckermr, line 96:
```suggestion
    val = np.uint32(val)
```

I suspect this might be simpler and the quickest.

arunkannawadi: You said that this was not safe.
https://github.com/esheldon/ngmix/pull/248#discussion_r2007582714

arunkannawadi: In practice, this saves you ~0.2 microseconds.

beckermr: I forget what was going on. Sorry for my confusion. Here is what happens 

```python
In [10]: v = np.uint32(2**32)
---------------------------------------------------------------------------
OverflowError                             Traceback (most recent call last)
Cell In[10], line 1
----> 1 v = np.uint32(2**32)

OverflowError: Python integer 4294967296 out of bounds for uint32

In [11]: v = np.uint32(2**31)

In [12]: np.asarray([2**32], dtype=np.uint32)[0]
---------------------------------------------------------------------------
OverflowError                             Traceback (most recent call last)
Cell In[12], line 1
----> 1 np.asarray([2**32], dtype=np.uint32)[0]

OverflowError: Python integer 4294967296 out of bounds for uint32

In [13]: np.asarray([2**31], dtype=np.uint32)[0]
Out[13]: np.uint32(2147483648)
```

beckermr: So in numpy 2, there is no difference in whether or not this kind of operation will raise an error. There may not have been in numpy 1 either.

We probably should be handling up to uint64 anyways, but that is an issue for another day.

arunkannawadi: RIght, and given that we don't consider the MSB, a direct `np.uint32` should be fine.

esheldon:
Are there any additional tests we should add to make sure we are comfortable?

arunkannawadi:
I had added a test against various integer types and lengths. As long as any of the implementations don't break that test, I think we can be comfortable. Please expand that test if you'd like and happy to go with any implementation of casting as long as the per-call time is under 10 microseconds. If you are willing to drop the support for `val` to be a 8-bit or 16-bit (unsigned) ints, a casting is not even necessary.

arunkannawadi:
https://github.com/esheldon/ngmix/blob/master/ngmix/tests/test_flags.py

beckermr:
I'm actually confused as to how this PR passes. I am going to try some stuff locally.

arunkannawadi:
FWIW, the CI uses numpy>=2, so the breakage you saw locally with numpy v1 wouldn't occur. Since you want to keep the support for numpy 1, I believe we should drop the last commit and go with the array based thing that this PR originally had. But please try this out locally and make any changes you'd like; you should have edit access to directly add commits here.

beckermr:
OK. Our previous test missed an important case of python ints. I have now rearranged things to explicitly truncate the extra bits.

beckermr:
We may want to emit a warning in this case or simply have the code error instead of truncating.

What do you think @esheldon?

ngmix/flags.py
```diff
@@ -84,6 +84,8 @@ def get_flags_str(val, name_map=None):
     # Cast to uint32 is sufficient given the range of flags.
     # This is because the second argument to the bitwise AND operator (&) below
     # would get implicitly cast to the same type as `val`.
+    if val >= 2**32:
+        val = val & 0xFFFFFFFF
```
esheldon, line 95: Can you make a comment how this works for future readers?

esheldon (commented):
I'd also be fine with just crashing if it is out of range.

ngmix/flags.py
beckermr: done!

ngmix/flags.py
```diff
@@ -84,14 +85,23 @@ def get_flags_str(val, name_map=None):
...
-    val = np.array(val, dtype=np.uint32)
+    if isinstance(val, int):
+        # Python 3.9 doesn't like to compute bitwise & with uint64 types,
+        # so we only do this for python ints where it is needed.
+        # The value 0xFFFFFFFF is an int with bits 0 to 31 (inclusive)
+        # set to 1 and the rest set to 0.
+        # The bitwise & operation with this mask sets all bits 32 and higher
+        # to zero. Thus the resulting int will fit into a unit32 numpy type.
+        val = val & 0xFFFFFFFF
+    val = np.asarray([val], dtype=np.uint32)
```
arunkannawadi:
```suggestion
    val = np.asarray([val], dtype=np.uint32)[0]
```

beckermr: thanks!

ngmix/flags.py
```diff
@@ -93,7 +93,7 @@ def get_flags_str(val, name_map=None):
         # The bitwise & operation with this mask sets all bits 32 and higher
         # to zero. Thus the resulting int will fit into a unit32 numpy type.
         val = val & 0xFFFFFFFF
-    val = np.asarray([val], dtype=np.uint32)[0]
+    val = np.asarray(val, dtype=np.uint32)
```
arunkannawadi: Huh? We're back to square one, the thing that causes the slowdown in the first place.

beckermr: Sorry hacking!

beckermr: The difference in "asarray" vs "array" is memory allocation done with array.

beckermr:
@arunkannawadi can you time this to see if it is still faster?

beckermr:
My machine is being odd on timing and I see no difference amongst the various options.

arunkannawadi:
This is slow (50 microseconds). A good rule-of-thumb is after casting `val` must not be of type array.

ngmix/flags.py
```diff
@@ -84,14 +85,23 @@ def get_flags_str(val, name_map=None):
...
-    val = np.array(val, dtype=np.uint32)
+    if isinstance(val, int):
+        # Python 3.9 doesn't like to compute bitwise & with uint64 types,
+        # so we only do this for python ints where it is needed.
+        # The value 0xFFFFFFFF is an int with bits 0 to 31 (inclusive)
+        # set to 1 and the rest set to 0.
+        # The bitwise & operation with this mask sets all bits 32 and higher
+        # to zero. Thus the resulting int will fit into a unit32 numpy type.
+        val = val & 0xFFFFFFFF
+    val = np.asarray(val, dtype=np.uint32)[0]
```
beckermr:
```suggestion
    val = np.asarray([val], dtype=np.uint32)[0]
```

ngmix/flags.py
arunkannawadi: Btw, this conversion is not the expensive part. It is the bitwise AND with an object of type `array` that is slow.

beckermr:
Thanks. Try the latest commit.

arunkannawadi:
This is good. 6 microseconds or so.

beckermr (approved):

ngmix/flags.py
```diff
@@ -84,14 +85,23 @@ def get_flags_str(val, name_map=None):
     # Cast to uint32 is sufficient given the range of flags.
     # This is because the second argument to the bitwise AND operator (&) below
     # would get implicitly cast to the same type as `val`.
-    val = np.array(val, dtype=np.uint32)
+    if isinstance(val, int):
```
arunkannawadi, line 88: I don't quite follow the logic for special casing here. If you cast it to `np.uint32` regardless, anything that's  not in the last 32 bits get dropped anyway. The casting works fine as is even in python 3.9. What does this gain?

arunkannawadi: *by casting, I mean the line after this `if` block.

beckermr: This if block handles cases where a pure python int is passed to the code.

Consider the following tests:

```
(py39) beckermr@mac ~ % ipy                  
Python 3.9.23 | packaged by conda-forge | (main, Jun  4 2025, 18:02:02) 
Type 'copyright', 'credits' or 'license' for more information
IPython 8.18.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import numpy as np

In [2]: np.__version__
Out[2]: '2.0.2'

In [3]: np.asarray([2**45], dtype=np.uint32)
---------------------------------------------------------------------------
OverflowError                             Traceback (most recent call last)
Cell In[3], line 1
----> 1 np.asarray([2**45], dtype=np.uint32)

OverflowError: Python integer 35184372088832 out of bounds for uint32
```

```
(py39) beckermr@mac ~ % ipy                  
Python 3.9.23 | packaged by conda-forge | (main, Jun  4 2025, 18:02:02) 
Type 'copyright', 'credits' or 'license' for more information
IPython 8.18.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import numpy as np
^[[A^[[A^[[A
In [2]: np.__version__
Out[2]: '1.26.4'

In [3]: np.uint64(2**45) & 0xFFFFFFFF
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[3], line 1
----> 1 np.uint64(2**45) & 0xFFFFFFFF

TypeError: ufunc 'bitwise_and' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''

In [4]: np.asarray([2**45], dtype=np.uint32)
<ipython-input-4-887e12c4c114>:1: DeprecationWarning: NumPy will stop allowing conversion of out-of-bound Python integers to integer arrays.  The conversion of 35184372088832 to uint32 will fail in the future.
For the old behavior, usually:
    np.array(value).astype(dtype)
will give the desired result (the cast overflows).
  np.asarray([2**45], dtype=np.uint32)
Out[4]: array([0], dtype=uint32)
```

I would just run the `&` operation for all types, but some numpy versions barf on that.


```
(py39) beckermr@mac ~ % ipy
Python 3.9.23 | packaged by conda-forge | (main, Jun  4 2025, 18:02:02) 
Type 'copyright', 'credits' or 'license' for more information
IPython 8.18.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import numpy as np

In [2]: np.__version__
Out[2]: '2.0.2'

In [3]: np.uint64(2**45) & 0xFFFFFFFF
Out[3]: np.uint64(0)

```

```
(py39) beckermr@mac ~ % ipy                  
Python 3.9.23 | packaged by conda-forge | (main, Jun  4 2025, 18:02:02) 
Type 'copyright', 'credits' or 'license' for more information
IPython 8.18.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import numpy as np
^[[A^[[A^[[A
In [2]: np.__version__
Out[2]: '1.26.4'

In [3]: np.uint64(2**45) & 0xFFFFFFFF
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[3], line 1
----> 1 np.uint64(2**45) & 0xFFFFFFFF

TypeError: ufunc 'bitwise_and' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''

```

So truncating python ints before putting them through the array is the best compromise I could find.

arunkannawadi: Ah, okay! There are so many combinations between python versions and numpy versions and I agree, this seems to cover all bases. I wonder if we need to have `numpy<2` in CI explicitly.

beckermr: We should have that in CI yes. I can do that in a separate PR.

beckermr: I added it here.

arunkannawadi:
I believe there's going to be a new `rubinenv` coming soon. if you can cut a release with this and other changes, we can get them in there.

Issue #255, opened 2026-04-23
Using a galsim object for a PSF instead of a Gaussian Mixture


s-Sayan:
Hi Erin,

First of all, thank you for developing such a flexible and generalizable codebase. I’ve implemented and validated ngmix within the SuperBIT pipeline ([link](https://github.com/superbit-collaboration/superbit-lensing/blob/main/superbit_lensing/metacalibration/ngmix_fit.py#L284-L342)), where we currently use the `em5` model for PSF fitting. One example of the fit is shown below.

<img width="564" height="251" alt="Image" src="https://github.com/user-attachments/assets/d837fc25-7f7e-420d-98aa-4b9c227a6414" />

My main concern is related to PSF modeling with a gaussian mixture. Since SuperBIT is a balloon-borne experiment, we don’t have atmospheric seeing contributing to PSF blurring. The PSF is not undersampled (pixel scale ~0.141", PSF FWHM ~0.47", the stability was an issue and the telescope jitter was much higher than we thought). We perform metacalibration using the psf="dilate" option (as you previously suggested) and include PSF shear terms to correct additive biases. However, it still seems that the Gaussian mixture approximation may not capture important PSF features, particularly for exposures with highly elliptical or unstable PSFs. In those cases, we observe PSF-to-galaxy ellipticity leakage and have had to discard some exposures for metacalibration (see our recent paper: [Lensing in the Blue III](https://arxiv.org/pdf/2603.18376)).

Eric Huff mentioned that ngmix may support using a GalSim object directly for the PSF, instead of a Gaussian mixture approximation. Although this would likely be slower than em5, the relatively small number of objects per SuperBIT target makes it potentially feasible for us.

I tried to locate this functionality in the codebase but wasn’t able to find it. If this option exists, could you please point me to how to use it (or whether it requires a custom implementation)?

Thanks again for your help, and for all your work on ngmix!

Best regards,
Sayan

esheldon:
You can use a galsim fitter with a real PSF image.    See ngmix.fitting.galsim_fitters

s-Sayan:
Hi Erin,

Thanks a lot for the guidance! I am trying to use `ngmix.fitting.GalsimFitter` in the `MetacalBootstrapper`, but I'm not sure I'm doing it right — one quick hint is that it runs much faster than the EM fitter I've been using (but my understanding is it should be much slower).

My understanding is that I don't need to set `psf_runner` anymore. Could you please check the following implementation and let me know if it's correct?

[...]

PR #259, opened 2026-06-10
added examples/fitting/fitting_galsimfitter.py & examples/metacal/metacal_psf_response.py

s-Sayan:
### New files
- **`examples/fitting/fitting_galsimfitter.py`** — Demonstrates fitting with `GalsimFitter`, analogous to the existing fitter examples in the same directory.
- **`examples/metacal/metacal_psf_response.py`** — Demonstrates metacalibration using `GalsimFitter` with `psf='dilate'`, including PSF response corrections. Follows the structure of `examples/metacal/metacal.py`.

### Notes 
Both examples are heavily inspired by the discussion in [issue #255](https://github.com/esheldon/ngmix/issues/255).

### DIFF
--- a/examples/fitting/fitting_galsimfitter.py
+++ b/examples/fitting/fitting_galsimfitter.py
@@ -1,0 +1,222 @@
+"""
+example fitting of an exponential model, using GalsimFiiter
+
+This example uses GalsimFitter inside a runner, 
+which directly uses the galsim GSOobject for the PSF to infer the Pre-PSF galaxy properties
+One quick note, GalsimFitter resdict['pars'][4] returns half light radius, (unlike fitter, which returns the size T) 
+
+A run of the code should produce output something like thid
+
+    > python fitting_galsimfitter.py
+
+    S/N: 805.3598771396064
+    true flux: 100 meas flux: 99.9136 +/- 0.427817 (99.7% conf)
+    true hlr: 0.5 meas hlr: 0.501831 +/- 0.0037977 (99.7% conf)
+    true g1: 0.05 meas g1: 0.0456032 +/- 0.00703185 (99.7% conf)
+    true g2: -0.02 meas g2: -0.022934 +/- 0.00702158 (99.7% conf)
+
+"""
+import numpy as np
+import galsim
+import ngmix
+
+lm_pars = {'maxfev':2000, 'xtol':5.0e-5, 'ftol':5.0e-5}
+
+
+def main():
+
+    args = get_args()
+    rng = np.random.RandomState(args.seed)
+
+    obs, obj_pars = make_data(rng=rng, noise=args.noise)
+
+    # fit the object to an exponential disk
+    prior = get_prior(rng=rng, scale=obs.jacobian.scale)
+    # fit using the levenberg marquards algorithm
+    fitter = ngmix.fitting.GalsimFitter(model='exp', prior=prior, fit_pars=lm_pars)
+    # make parameter guesses based on a psf flux and a rough T
+    gm_guess = ngmix.gaussmom.GaussMom(1.2).go(obs)
+    guesser = ngmix.guessers.TFluxAndPriorGuesser(rng=rng, T=gm_guess["T"], flux=gm_guess['flux'], prior=prior)
+
+    runner = ngmix.runners.Runner(fitter=fitter, guesser=guesser, ntry=20)
+
+    boot = ngmix.bootstrap.Bootstrapper(
+        runner=runner,
+        psf_runner=None,
+    )
+
+    res = boot.go(obs)
+
+    print()
+    print('S/N:', res['s2n_r'])
+    print('true flux: %g meas flux: %g +/- %g (99.7%% conf)' % (
+        obj_pars['flux'], res['flux'], res['flux_err']*3,
+    ))
+    print('true hlr: %g meas hlr: %g +/- %g (99.7%% conf)' % (
+        obj_pars['hlr'], res['pars'][4], res['pars_err'][4]*3,
+    ))
+    print('true g1: %g meas g1: %g +/- %g (99.7%% conf)' % (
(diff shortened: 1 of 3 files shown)


esheldon:
Thanks for the examples.  Do you think a unit test could also be made for these?

s-Sayan:
Hi Erin, yes — these are unit tests (similar to existing examples in your directory: `fitting_galsimfitter.py` mirrors `examples/fitting/fitting.py`, and `metacal_psf_response.py` mirrors `examples/metacal/metacal.py`). They use round objects, a Moffat PSF (with slight PSF shears), and low-noise settings. Running the code should produce the following outputs:

**`fitting_galsimfitter.py`**
```text
S/N: 805.3598771396064
true flux: 100   meas flux:  99.9136 +/- 0.427817  (99.7% conf)
true hlr:  0.5   meas hlr:   0.501831 +/- 0.0037977 (99.7% conf)
true g1:   0.05  meas g1:    0.0456032 +/- 0.00703185 (99.7% conf)
true g2:  -0.02  meas g2:   -0.022934 +/- 0.00702158 (99.7% conf)
```

**`metacal_psf_response.py`** *(takes a bit of time — runs over 5 input shears: −0.02, −0.01, 0.0, 0.01, 0.02, then fits a line to obtain m and c)*
```text
S/N: 50976.7
R11: 0.999596
R11_psf: 2.65024e-05
m: 0.000421002 +/- 0.000118555 (99.7% conf)
c: 1.21787e-07 +/- 1.66119e-06 (99.7% conf)
```

I've added explanations and expected outputs to the docstrings at the top of each file. Is there any other test in particular you'd like me to add?

esheldon:
I meant real unit tests that assert correctness, as in ngmix/tests/

but never mind that for now.  I'll see if these run for me when I get a chance.

s-Sayan:
Ah I see, thanks! I didn't know about `pytest` — that's pretty cool!

I can add something like that for testing the `GalsimFitter` in `MetacalBootstrapper` and the PSF response correction, and assert correctness for the additive `c` (something similar to `ngmix/tests/test_metacal_accuracy.py`). I guess I would only test this for `psf='dilate'` — is there any specific situation you would suggest testing? Like any weird PSF maybe?

I also see you already have a unit test for the `GalsimFitter`: `ngmix/tests/test_ml_fitting_galsim.py::test_ml_fitting_galsim`.

s-Sayan:
Okay I added a unit test at `ngmix/tests/test_metacal_galsim_psf_response.py`. To make the test a bit more challenging, I used `galsim.OpticalPSF` to create a jittery PSF (using SuperBIT instrumentation details) and shear the PSF a bit, since I cannot use a `.psf` file, I thought this is the best I can do, here one such example of what that PSF looks like:

image

I check the additive bias c1 (using g1 component only) for different PSF ellipticity combinations (`psf_g1 = -0.02, +0.02`; `psf_g2 = -0.02, +0.02`) and assert correctness via `cerr < 1e-5` and `|c| < cerr` (consistent with zero). Let me know if you'd like a different assertion condition or additional scenarios to be tested. The values I get for those four combinations are:

| `psf_g1` | `psf_g2` | `c1`        | `3*cerr` (99.7%)  |
|----------|----------|------------|-----------------|
| −0.02   | −0.02    | 6.24×10⁻⁷  | 7.63×10⁻⁶       |
| +0.02    | −0.02    | 4.09×10⁻⁷  | 7.63×10⁻⁶       |
| −0.02    | +0.02    | 7.37×10⁻⁷  | 7.62×10⁻⁶       |
| +0.02    | +0.02    | 5.41×10⁻⁷  | 7.61×10⁻⁶       |

P.S. I have also modified `examples/metacal/metacal_psf_response.py` to use the SuperBIT optical PSF with shear rather than Moffat PSF, and updated the expected output in the docstring. For both the unit test and the example, shear estimation uses only `sheear_true = ±0.02` to estimate `m` and `c`, with `ntrial=100` to keep things a bit faster.

esheldon:
I don't think you would always expect c < cerr, so this unit test can fail (it did for me)

esheldon:
These would be good checks I think
```
abs(m) < 0.001
abs(c) < 1.0e-5
```
I found setting the galaxy size to something smaller also helped (I used 128).

Also I ran tests with `psf="azgauss"` it it looks good too.

s-Sayan:
Thanks! I have now updated the assertion conditions like that, in the same code I have also included the `psf=azgauss` settings to be tested with the superbit optical psf.

> I found setting the galaxy size to something smaller also helped (I used 128).

I think I understand this but just to make sure — I use an `Exponential` profile with `hlr=0.5` arcsec (with pixel scale 0.141). You mean setting `nx, ny = 128` while doing:

​```
im = obj.drawImage(nx=npix, ny=npix, scale=scale).array
​```

Previously I was letting GalSim decide `npix`, and that was actually causing higher `m` and `c`, hence the failure for `azgauss`.

**TL;DR:** Is it a rule of thumb that `psf_npix` shouldn't be much higher than `npix` of the object? Or is it specific to the `psf='azgauss'` settings?

s-Sayan:
Hey Erin, I have a quick question on how are we tracking if the `psf='azgauss'`, is better? is it just looking at m and c. If so, then the difference between the `psf='azgauss'` and `psf='gauss'` is not so much (with the superbit psf test above), I am getting the c values small for both the settings.
1. `psf='azgauss'`:
```bash
m: 0.00042347 +/- 0.000426856 (99.7% conf)
c: -1.71905e-06 +/- 8.53713e-06 (99.7% conf)
```
2. `psf='gauss'`:
```bash
m: 0.000431744 +/- 0.00042775 (99.7% conf)
c: -1.82156e-06 +/- 8.555e-06 (99.7% conf)
```
I have also tried crancking up the noise, the m and c values are higher, but they are comparable for the both settings. Maybe I should try some other test to see the improvement over `gauss`?

esheldon:
azgauss is better when the PSF image is noisy

PR #260, opened 2026-06-11
Add a noise-robust version of psf='gauss' for metacal

esheldon:
This version is the same algorithm as Mike's original, but instead of looking at individual pixels, which are noisy, it makes an azimuthally averaged radial profile.

The constants have also been adjusted to reduce noise amplification at mid-k


esheldon:
I also factored out the the fitgauss algorithm for testing purposes

esheldon:
gauss_method_noise_test_ellip

esheldon:
Power ratio for real PSF.  The new method suppresses high k power, and amplifies mid-k power less than fitgauss or original gauss

power_ratio_profile_r_11_11

esheldon:
this method does something reasonable for various types of profiles

test_profiles

beckermr (changes requested):
The code looks great.

However, this change is breaking in that the meaning of `gauss` when used with metadetect has changed.

Can we instead add a new name for the new algorithm? Maybe `gaussprof` or something?

esheldon:
It breaks unit tests that expect an exact answer.  Otherwise it is a strict improvement.  Would it break anything other than unit tests?

beckermr:
It breaks reproducibility for downstream users of the code.

esheldon:
Yeah, that's like a test.  I hear you for sure, but there are times when it makes sense to break.

Right now 'gauss' is our default and it doesn't work well for noisy psfs.  We need people to be able to use the defaults and get something that works well.  the only way to accomplish that is to change the default.

A different route:  We could make it really break:  make psf= a required keyword rather than have a default.  Current codes using the default crash with a loud message saying "send psf=".   We remove 'gauss' as an option.

beckermr:
Agreed that it makes sense to use the new thing. Here are other options:

1. Bump the ngmix major version to preserve semver.
2. Declare a deprecation process (say, adding a deprecation warning, waiting XYZ amount of time, and then a change on the next minor version).

esheldon:
If we deprecate, should it be "the behavior of psf='gauss' will change" or "the default will change to..."?

beckermr:
I don't care between those two.

beckermr:
I do not think it is good policy to remove a version of something people used in the past for scientific work possibly.

esheldon:
OK, some possible options, other ideas?

1. warning:  In the next ngmix version (xxx) psf='gauss' will use an improved algorithm, which will result in small changes in the metacal images.
2. Same but give them a way to use the new one now with a function call, e.g. `ngmix.metacal.use_improved_psf_gauss()`. Becomes a no-op in the future.
3. warning: In the next ngmix version (xxx) the default psf= setting will change from 'gauss' to `whatever_we_call_it`

beckermr:
Let's go with 3.

esheldon:
renamed to `psf='azgauss'`

ngmix/metacal/azgauss_target_psf.py
```diff
@@ -0,0 +1,130 @@
+"""
+Noise-robust derivation of the round gaussian target reconvolution PSF
+for metacal.
+
+Drop-in replacement for metacal._get_gauss_target_psf (originally from galsim
+tests/test_metacal.py).  That version finds the pinning scale with a pixelwise
```
beckermr, line 12: What does "pinning scale" mean here?

esheldon:
Shall we merge?

beckermr (commented):
One comment on language in the docs I did not understand. Otherwise LGTM!

ngmix/metacal/azgauss_target_psf.py
esheldon: That's the scale where the power drops below some threshold.  That's where you "pin" the reconvolution kernel to be less than some factor of the original PSFs power.

beckermr (approved):
LGTM! Thank you!

PR #262, opened 2026-07-18
allow sending a noise image to prepsfmom fitters

esheldon:
If the user sets use_noise_image=True, noise power for the error estimates is measured from the noise image attached to the observation using |fft(zero padded noise)|^2.  It is also boosted by eff_pad_factor^2

This is useful when the noise is correlated.  by working in k space the mode powers can be treated independently to give the correct variance scaling

The error estimates work well.  in a new unit test with correlated noise the flux scatter and error estimate matches to the percent level, while the old version underestimtes it by a factor of 4.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,13 @@
 
 ### New features
 
+    - The prepsfmom fitters (KSigmaMom, PGaussMom) accept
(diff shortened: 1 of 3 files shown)


beckermr (approved):
LGTM! Nice work!

PR #265, opened 2026-07-30
Analytic derivatives for LM noise

esheldon:
Improve the speed of the noise_cov using analytic derivatives

    Use analytic derivatives for the simple models (gauss, exp, dev)
    and fall back to finite deriv. for the complex models (bulge disk)

    A new numba kernel is added to keep it fast.

    This is about 5.6 times faster than the FD version.


esheldon:
All tests still pass, behavior unchanged, just faster.   Going to merge.

PR #267, opened 2026-07-30
Update the fast exp to be twice differentiable

esheldon:
replace the fast exponential (fexp), which had steps in its derivatives, with a standard form based on Chebyshev polynomials.

Same speed, but 17 times more accurate and twice differentiable.

This fixes some issues we had comparing the analytical derivatives with the finite diff derivatives, but should otherwise not affect the fits in any noticeable way.  The old exp5 is still available but no longer aliased to fexp


beckermr (approved):

esheldon:
trying to figure out these flaky failures

esheldon:
Its because with the new fastexp the fits are more stable, so the assertion of failure never happens

PR #271, opened 2026-08-19
Kspace admom

esheldon:
K space adaptive moments.

Works across multiple bands/epochs.

Produces prepsfmom like results, as it is inspired by that work.

Can fit gauss, which uses exactly adaptive moments, or exp/dev/bdf which use a trick to make the fixed point method work for the "family" of fixed relative gaussians.

This also includes a fix that was needed along the way, non-square support


esheldon:
the new numba `*_status` functions are to deal with the numba bug reported here:  https://github.com/numba/numba/issues/10783

beckermr:
@esheldon if you'd like code review here, it'd be very helpful to break this up into smaller PRs.

esheldon:
I hear you and I'll give it a try.  But it all flows together, and a lot of it won't work without the other parts.

beckermr:
I will take a quick look later and see if I can find some ways to help break it up.

beckermr:
For sure the `_status` stuff can be broken into its own PR. That won't help a ton, but it will break off a few hundred lines.

beckermr:
The change to allow normal prePSF moments to use non-square images can also be its own PR.

beckermr:
I'd also break out the Gaussian models from support for the exp/dev/bdf models.

esheldon:
this is the same test failure with too tight a tolerance that failed in PR #272 

will wait and merge that in

beckermr (commented):
There is still way too much code here for me to review. 

There are a few options:

1) We can break up the code further by separating out each of the models or the models versus the Gaussian into their own PRs 

2) I can only look at the tests, which is still a lot, but is more manageable. 

3) We can go without code review. I don't like this option, but at the end of the day this is your repo and  I won't insist on code review. 

I'm happy to work on 1), but my time is limited, and I have not gotten to it yet. It'd be helpful if you can turn on the setting that offers to update PR branches in the repo settings on github. 

A small observation from my brief look already. Many of the doc strings are unintelligible (at least to me, but probably others). I get that there is some non-trivial math happening here, but I think they could be better.

ngmix/prepsfadmom/errors.py
```diff
@@ -104,39 +138,43 @@ def flux_var_delta(Sigma, sums, cov, fsums, fvars, fmcovs):
...
     full cross-band covariance of the raw per band flux sums,
-    including the first order response of the adaptive weight to
-    the noise (the delta method); the diagonal is flux_var_delta.
+    including the adaptive weight response; the diagonal is
+    flux_var_delta
 
     The fluctuation dF_b is proportional to
-    dS_Fb - r_b dS_F + r_b b . dS_M (see flux_var_delta): the
-    joint flux and moment sum responses are shared by all bands,
+    dS_Fb - r_b dS_F + r_b b . dS_M (see flux_var_delta).  The
```
beckermr, line 158: Where does this formula come from? What is `r_b`? The reference to `flux_var_delta` doesn't explain that.

ngmix/prepsfadmom/errors.py
```diff
@@ -156,9 +194,10 @@ def flux_cov_delta(Sigma, sums, cov, fsums, fvars, fmcovs):
     g = fmcovs @ b - fvars
```
beckermr, line 194: This code needs more comments and explanations if you want anyone else to understand it.

ngmix/prepsfadmom/errors.py
```diff
@@ -0,0 +1,662 @@
...
+    s = cff + bcb - 2 * (b @ cmf)
+    g = fmcovs @ b - fvars
+
+    fcov = np.outer(r, r) * s + np.outer(r, g) + np.outer(g, r)
+    # the band flux-sum noises are independent across bands, so
+    # fvars enters only on the diagonal, written in the direct
+    # per-band variance form to keep it bitwise identical to the
+    # historic flux_var_delta
+    idx = np.arange(fsums.size)
+    fcov[idx, idx] = fvars + r ** 2 * s + 2 * r * g
```
beckermr, line 202: The function flux_var_delta above literally returns the diagonal of the output of this function. Thus this block should not be needed at all.

ngmix/prepsfadmom/errors.py
```diff
@@ -0,0 +1,662 @@
...
+
+        d(fam_cov) = B^-1 dM,    B = dM_pred/d(fam_cov)
+        dF_b/F_b = dfs_b/fs_b - (dln N/d(fam_cov)) . d(fam_cov)
+
+    where dM is the noise fluctuation of the measured moment ratios.
+    B and the normalization derivative are evaluated by central
+    differences on the closed form model predictions, so unlike the
+    gauss delta method there is no gaussian approximation beyond
+    replacing the data by the converged model; for a single gaussian
+    family this reduces exactly to flux_var_delta.
```
beckermr, line 360: This comment needs more explanation. What is `fs_b`?

Why isn't `dF_b/F_b = 1`?

ngmix/prepsfadmom/errors.py
```diff
@@ -0,0 +1,662 @@
...
+
+    Parameters
```
beckermr: Not all parameters are documented.

ngmix/prepsfadmom/errors.py
```diff
@@ -0,0 +1,662 @@
...
+    fcov_raw = (
+        np.outer(fsums, fsums) * aca
+        - np.outer(fsums, acmf) - np.outer(acmf, fsums)
+        + np.diag(fvars)
+    )
+    var_raw = np.diag(fcov_raw).copy()
+    return var_raw, fam_err_cov, fcov_raw
+
+
+def bdf_joint_sandwich(
```
beckermr, line 484: There is no way I can verify what this piece of code is doing without a lot more details on the math. They can either be embedded in comments in the code or this part of the code can go unreviewed.

ngmix/prepsfadmom/full_errors.py
```diff
@@ -0,0 +1,854 @@
...
+plane through the stored image->kim transfer (epoch['ktransfer'])
+and contracted with the per-pixel variance from the weight map.
+Apodization enters exactly: the masked prep's image->kim map
+factorizes as ktransfer(k) e^{-ik.x} m(x), so influence_kernels
+applies the mask in pixel space, where it is diagonal.  The
+kernel derivatives with respect to the weight covariance and
+center are also closed form, allowing the data sums to be
+linearized about a solution so that fixed-point Jacobians never
+touch the data modes.  This construction, including the exact
+apodization treatment, was derived for this code.
```
beckermr, line 19: This comment is clearly left over from Claude. Do you want that in the final code?

beckermr, line 19: This comment is clearly left over from Claude. Do you want that in the final code?

ngmix/prepsfadmom/full_errors.py
beckermr, line 1: This file is ~1k lines of extremely dense code, presumably produced with help from Claude?

I will need it to be commented line by line, explaining the steps in the derivation, in order to review it. Absent that, I can't.

ngmix/prepsfadmom/models.py
```diff
@@ -0,0 +1,260 @@
...
+        [T * (1 - e1), T * e2],
+        [T * e2, T * (1 + e1)],
+    ])
+
+
+def mom_from_cov(S):
+    """moments (M1, M2, T) = (<uu - vv>, 2<vu>, <vv + uu>) from a
+    covariance matrix [[Ivv, Ivu], [Ivu, Iuu]], the inverse of
+    cov_from_e up to the T normalization of the shapes"""
+    return S[1, 1] - S[0, 0], 2 * S[0, 1], S[0, 0] + S[1, 1]
```
beckermr, line 80: The docstring in the first function is not complete. Also, I would have expected these utility functions to be in `ngmix/moments.py`.

ngmix/prepsfadmom/models.py
```diff
@@ -0,0 +1,260 @@
...
+    # Sp: the second moments of the product gaussian
+    Sp = wt_cov @ Cinv @ cov_sm
+    vv = Sp[0, 0] + mu[0] * mu[0]
+    vu = Sp[0, 1] + mu[0] * mu[1]
+    uu = Sp[1, 1] + mu[1] * mu[1]
+
+    return sflux * np.array([
+        mu[0], mu[1], uu - vv, 2 * vu, uu + vv, 1.0,
+    ])
+
```
beckermr, line 124: The computation of sflux is not clear and needs comments.

The function needs comments on the various steps. 

It appears to be using the formulas from this link: https://math.stackexchange.com/a/964103

ngmix/prepsfadmom/models.py
```diff
@@ -0,0 +1,260 @@
...
+    """flux fractions and relative T factors of the ngmix gaussian
+    expansion of the named profile ('exp': 6 gaussians, 'dev': 10
+    gaussians), normalized to total T = 1"""
+    if name not in _PROFILE_COMPS:
+        gm = GMixModel([0, 0, 0, 0, 1.0, 1.0], name)
+        d = gm.get_data()
+        fracs = d['p'] / d['p'].sum()
+        cTs = d['irr'] + d['icc']
+        _PROFILE_COMPS[name] = list(zip(fracs, cTs))
+    return _PROFILE_COMPS[name]
```
beckermr, line 139: This function returns constants that are already defined explicitly in the code, certainly in the case of `p`. Why recompute things instead of simply pulling in the constants that are needed?

ngmix/prepsfadmom/models.py
```diff
@@ -0,0 +1,260 @@
...
+    """
+    return (
+        [((1 - fracdev) * f, cT)
+         for f, cT in get_profile_comps('exp')]
+        + [(fracdev * f, TdByTe * cT)
+           for f, cT in get_profile_comps('dev')]
+    )
+
+
+def _mixture_comps(model):
```
beckermr, line 157: Why is this function private, but the ones above are public?

ngmix/prepsfadmom/models.py
```diff
@@ -0,0 +1,260 @@
...
+    """
+    the (frac, cT) component table for a mixture model dict or
+    spec; see the module docstring
+    """
+    if model['type'] == 'bdf':
+        return bdf_comps(model['fracdev'], model['TdByTe'])
+    return get_profile_comps(model['type'])
+
+
+def model_comps(model, Tsmooth):
```
beckermr, line 167: This function name is pretty unclear.

ngmix/prepsfadmom/models_nb.py
```diff
@@ -0,0 +1,100 @@
...
+
+        Cd0 = Ci00 * dv[k] + Ci01 * du[k]
+        Cd1 = Ci01 * dv[k] + Ci11 * du[k]
+
+        sflux = F[k] * np.exp(
+            -0.5 * (dv[k] * Cd0 + du[k] * Cd1)
+        ) / (nrm * np.sqrt(det))
+
+        mu0 = sw00 * Cd0 + sw01 * Cd1
+        mu1 = sw01 * Cd0 + sw11 * Cd1
```
beckermr, line 79: This bit of code needs comments on what it is computing in each block. It is assembling the mean via the formula at the math exchange post from above.

ngmix/prepsfadmom/full_errors.py
beckermr, line 1: FWIW, JAX enables you to compute the gradients of a fixed-point iteration scheme using the implicit function theorem. See the docs here: https://docs.jax.dev/en/latest/301/custom-derivatives.html#example-implicit-function-differentiation-of-iterative-implementations

What this means is that once you write the iteration function to solve for the fit, JAX will almost automatically give you the Jacobian matrices you need to then compute the proper error covariance of the output.

ngmix/prepsfadmom/errors.py
esheldon: It is needed, it was a bad comment.  the diagonal of the outer-product expression alone omits the per-band flux-sum variances, which appear only on the diagonal because the band flux-sum noises are independent across bands.

The old comment made it read like a legacy/compatibility artifact, which was wrong

I'll update to make it clearer:  `fcov[idx, idx] +=fvars`

ngmix/prepsfadmom/models.py
esheldon: audit for private/public coming. Some things are needed outside the package.

ngmix/prepsfadmom/full_errors.py
esheldon: I added that for you so you would know there is no ref for it

PR #272, opened 2026-08-24
Work around numba memory leak on raise from jitted code

esheldon:
Split out of #271 at reviewer request.

The numba runtime leaks a small allocation on every exception raised from compiled code, and the fitters reject invalid parameters through these paths at high rate on hard data. This adds non-raising `_status` variants of the hot-path numba functions (`gauss2d_set_norm`, `gmix_set_norms`, `gmix_fill_*`, `g1g2_to_e1e2`, `fill_fdiff`) that return an error code, with `GMixRangeError` raised from python in `GMix._fill` and `FitModel`. The raising versions remain as thin wrappers around the `_status` versions, so the API is unchanged. If numba fixes the leak these can be removed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01PK1a41EiKVLGjyq3Bp6RGa

### DIFF
--- a/ngmix/fitting/results.py
+++ b/ngmix/fitting/results.py
@@ -5,7 +5,9 @@ from .. import gmix
 from ..gexceptions import GMixRangeError
 from ..defaults import PDEF, CDEF, LOWVAL, BIGVAL, copy_if_needed
 from ..observation import Observation, ObsList, get_mb_obs
-from ..gmix.gmix_nb import gmix_convolve_fill, fill_fdiff
+from ..gmix.gmix_nb import (
+    gmix_convolve_fill, fill_fdiff_status, GMIX_STATUS_MESSAGES,
+)
 from ..gmix import GMixList, MultiBandGMixList
 from ..flags import ZERO_DOF, DIV_ZERO, BAD_VAR
 
@@ -454,9 +456,16 @@ class FitModel(dict):
             start = self._fill_priors(pars=pars, fdiff=fdiff)
 
             for pixels, gm in zip(self._pixels_list, self._gmix_data_list):
-                fill_fdiff(
+                # status form: the numba runtime leaks per raise,
+                # and this call rejects invalid parameters at
+                # high rate on hard data (norms not settable)
+                status = fill_fdiff_status(
                     gm, pixels, fdiff, start,
                 )
+                if status != 0:
+                    raise GMixRangeError(
+                        GMIX_STATUS_MESSAGES[status]
+                    )
 
                 start += pixels.size
 
--- a/ngmix/fitting/results.py
+++ b/ngmix/fitting/results.py
@@ -6,7 +6,7 @@ from ..gexceptions import GMixRangeError
 from ..defaults import PDEF, CDEF, LOWVAL, BIGVAL, copy_if_needed
 from ..observation import Observation, ObsList, get_mb_obs
 from ..gmix.gmix_nb import (
-    gmix_convolve_fill, fill_fdiff_status, GMIX_STATUS_MESSAGES,
+    gmix_convolve_fill, fill_fdiff_status, get_status_message,
 )
 from ..gmix import GMixList, MultiBandGMixList
 from ..flags import ZERO_DOF, DIV_ZERO, BAD_VAR
@@ -464,7 +464,7 @@ class FitModel(dict):
                 )
                 if status != 0:
                     raise GMixRangeError(
-                        GMIX_STATUS_MESSAGES[status]
+                        get_status_message(status)
                     )
 
                 start += pixels.size
--- a/ngmix/gmix/gmix_nb.py
+++ b/ngmix/gmix/gmix_nb.py
@@ -187,6 +187,42 @@ GMIX_STATUS_MESSAGES = {
 }
 
 
+def get_status_message(status):
+    """
+    the message for a gmix status code (GMIX_STATUS_MESSAGES),
+    with a fallback for unknown codes.  All python-level raises
+    from status codes go through this function
+    """
+    return GMIX_STATUS_MESSAGES.get(
+        status, "unknown gmix status %d" % status,
+    )
+
+
+@njit
+def gmix_status_raise(status):
+    """
+    raise GMixRangeError for a nonzero status code; no-op for
+    zero.  All jitted raises from status codes go through this
+    function.
+
+    numba requires exception messages to be compile-time
+    constants, so this function cannot look the message up in
+    GMIX_STATUS_MESSAGES at runtime; the branches below MUST
+    mirror that dict (enforced by test_gmix_status_messages),
+    and any new code needs an entry in both places
+    """
+    if status == 0:
+        return
+    elif status == 1:
+        raise GMixRangeError("g >= 1")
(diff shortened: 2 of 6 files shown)


esheldon:
I used claude code to split this out from #271

ngmix/gmix/gmix_nb.py
```diff
@@ -675,6 +787,17 @@ def g1g2_to_e1e2(g1, g2):
...
+
+
+@njit
+def g1g2_to_e1e2(g1, g2):
+    """
+    convert g to e
+    """
+    e1, e2, status = g1g2_to_e1e2_status(g1, g2)
+    if status != 0:
+        raise GMixRangeError("g >= 1")
```
beckermr: This bit of code should use the global status message dict above. The logic is that if the status codes change, this function should change too.

ngmix/gmix/gmix_nb.py
```diff
@@ -890,14 +1015,38 @@ def fill_fdiff(gmix, pixels, fdiff, start):
...
+    pixels: array if pixel structs
+        u,v,val,ierr
+    fdiff: array
+        Array to fill, should be same length as pixels
+    """
+    status = fill_fdiff_status(gmix, pixels, fdiff, start)
+    if status == 2:
+        raise GMixRangeError("det too low")
+    elif status == 3:
+        raise GMixRangeError("T too low")
```
beckermr: Same here for using the global status dict above.

ngmix/gmix/gmix_nb.py
```diff
@@ -338,6 +386,19 @@ def gmix_fill_simple(gmix, pars, fvals, pvals):
...
+@njit
+def gmix_fill_simple(gmix, pars, fvals, pvals):
+    """
+    fill a simple (6 parameter) gaussian mixture model
+
+    no error checking done here
+    """
+    status = gmix_fill_simple_status(gmix, pars, fvals, pvals)
+    if status != 0:
+        raise GMixRangeError("g >= 1")
```
beckermr: Same here for using the global status dict. Right now the code is ignoring the status number. What happens if status != 1?

ngmix/gmix/gmix_nb.py
```diff
@@ -216,6 +242,26 @@ def gauss2d_set_norm(gauss):
...
+    parameters
+    ----------
+    gauss: a 2-d gaussian structure
+        See gmix.py
+    """
+    status = gauss2d_set_norm_status(gauss)
+    if status == 2:
+        raise GMixRangeError("det too low")
+    elif status == 3:
+        raise GMixRangeError("T too low")
```
beckermr: Same here for the global status dict.

ngmix/gmix/gmix_nb.py
```diff
@@ -183,29 +212,26 @@ def gmix_set_norms(gmix):
     gmix:
        gaussian mixture
     """
-    for gauss in gmix:
-        gauss2d_set_norm(gauss)
+    status = gmix_set_norms_status(gmix)
+    if status == 2:
+        raise GMixRangeError("det too low")
+    elif status == 3:
+        raise GMixRangeError("T too low")
```
beckermr: Same here for the global status dict.

beckermr (changes requested):
The status codes need to always be passed through the global status code to message function.

ngmix/gmix/gmix_nb.py
esheldon: numba requires exception messages to be compile time known

esheldon: the best I can do is make sure they are the same in both places

beckermr: Raising errors from numba was the source of the leak in the first place. So this function should be split into its `_status` version with `njit` and a non-JITed version that raises, right?

esheldon: I didn't follow.  the original versions still exist and raise, but we don't use them and won't until the bug fix is accepted.

esheldon: the "non jit" version is using `get_status_message` in the fitters and then raising there, this is being done (see results.py)

ngmix/gmix/gmix_nb.py
```diff
@@ -173,6 +173,71 @@ def gmix_get_e1e2T(gmix):
...
+    GMIX_STATUS_MESSAGES at runtime; the branches below MUST
+    mirror that dict (enforced by test_gmix_status_messages),
+    and any new code needs an entry in both places
+    """
+    if status == 0:
+        return
+    elif status == 1:
+        raise GMixRangeError("g >= 1")
+    elif status == 2:
+        raise GMixRangeError("det too low")
```
beckermr, line 219: I still do not understand this PR. The original description says that numba leaks memory when exceptions are raised inside JITed functions, and yet this function is decorated with `njit` and raises an exception. What am I missing?

beckermr (changes requested):

ngmix/gmix/gmix_nb.py
esheldon: I just did this refactor so it would be clear to you that the strings used here match the ones in the dict.  No behavior changed

esheldon: Your original request cannot be achieved, it cannot use the dict.  numba only allows raising compile time known strings

ngmix/fitting/results.py
```diff
@@ -454,9 +456,16 @@ def calc_fdiff(self, pars):
...
             for pixels, gm in zip(self._pixels_list, self._gmix_data_list):
-                fill_fdiff(
+                # status form: the numba runtime leaks per raise,
+                # and this call rejects invalid parameters at
+                # high rate on hard data (norms not settable)
+                status = fill_fdiff_status(
                     gm, pixels, fdiff, start,
                 )
+                if status != 0:
+                    raise GMixRangeError(
```
esheldon, line 466: See here

ngmix/gmix/gmix_nb.py
esheldon: As for how these functions are actually used, and how this gets around the bug, the `_raise` versions are used from the fitting code (I added  a comment "see here")  and the raise is done from regular python.

esheldon: sorry I meant to write `_status` versions

beckermr: Sure but memory will be leaked anywhere a raise is used in JIT. Don't you need to remove all of that to work around the bug?

esheldon: Yes, that code does not get around the bug, it would leak.  That is why the `_status` versions were added and are now used instead of the raising codes. The raising codes are still there 1) so the api does not break and 2) we can use them in the future when the bug is fixed.

I added that specific code so you could see that the strings matched the dict, which I thought was your original worry that these would get out of sync.

beckermr (approved):
Ok. One could argue the api doesn't include whether or not the function is jitted, but let's not do that. 

Lgtm and thanks for the extra effort!

esheldon:
thanks for the review

PR #273, opened 2026-08-24
Allow non-square images in the prepsfmom fitters

esheldon:
Split out of #271 at reviewer request.

The pre-PSF moment fitters (`KSigmaMom`, `PGaussMom`) previously raised on non-square images. They now zero pad to square internally: the target dimension is the largest dimension of the image and psf times `pad_factor`, and the effective pad factor used to scale the noise is the ratio of padded to unpadded pixel counts. Apodization still works correctly. `_zero_pad_image` now pads each axis separately and returns the row and column offsets.

Tests cover trimming either axis and check results against the square case.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01PK1a41EiKVLGjyq3Bp6RGa

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -27,6 +27,8 @@
       differentiable in the parameters everywhere; the window costs
       about a percent in the pixel loop and removes 1.5e-5 of a round
       gaussian's flux where the cut removed 3.7e-6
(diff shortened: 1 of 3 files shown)


esheldon:
Claude code was used to split this from #271

beckermr (approved):

